PR #4, opened 2016-05-26
store files of disk instead of leveldb

juliangruber:
cc @mafintosh

### DIFF
--- a/app.js
+++ b/app.js
@@ -11,20 +11,22 @@ const fileReader = require('filereader-stream');
 const fs = require('fs');
 const {basename} = require('path');
 const raf = require('random-access-file');
+const downloadsFolder = require('downloads-folder')();
 
 const appPath = `${app.getPath('appData')}/${app.getName()}`;
-try { fs.mkdirSync(`${appPath}/files`) } catch (_) {}
+const filesPath = `${downloadsFolder}/dat`;
+try { fs.mkdirSync(filesPath) } catch (_) {}
+const keyPath = `${appPath}/key.txt`;
 
 const db = level(`${appPath}/db`);
 const drive = hyperdrive(db);
 
 let key;
-const keyPath = `${appPath}/key.txt`;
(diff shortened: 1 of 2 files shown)


app.js
```diff
@@ -18,7 +20,10 @@ let key;
 const keyPath = `${appPath}/key.txt`;
 try { key = fs.readFileSync(keyPath) } catch (_) {}
 
-const archive = drive.createArchive(key, { live: true });
+const archive = drive.createArchive(key, {
+  live: true,
+  file: name => raf(`${appPath}/files/${name}`)
```
mafintosh: i think we should put them in the downloads folder (whatever the equivalent is on each os). this points to the electron folder right?

mafintosh: maybe @feross has a module that returns the downloads folder for every os? webtorrent probably does something similar

juliangruber: I'm not sure putting it in the downloads folder is the right idea. When you have limited disk like I do you frequently purge it. On the other hand, that's exactly the right thing if you just use this for experimentation for now - then the data you fetch is short lived anyways.

> this points to the electron folder right?

Yup, that's in `Application Support`.

Hm still `~/Downloads` is probably the best folder. A quick npm search didn't reveal any module for it.

What do others think?

maxogden: +1 ~/Downloads folder for Mac. not sure about x-platform. I think `home-dir` is the module for getting the users home dir. I think the folder should be easy to change eventually but for now Downloads seems like a sensible default

juliangruber:
@mafintosh updated

mafintosh:
@juliangruber looks great! 👍

feross:
We used to use an approach like @juliangruber's module, but some advanced users on Windows complained that they changed the location of their Downloads folder in the OS settings.

To get around that, we just used Electron's [`app.getPath(name)`](https://github.com/electron/electron/blob/master/docs/api/app.md#appgetpathname)

```js
app.getPath('downloads')
```

PR #25, opened 2016-07-04
Styles

Kriesse:
### dat-design and Sass
Added dat-design and set up Sass, in order to re-use all the base styles we have in dat-design.
(using Sass instead of plain CSS allows us to access variables, mixins etc).

I noticed @laurengarcia's suggestion [here](https://github.com/datproject/design/issues/5#issuecomment-229448929) too late, but I think we chose almost the same approach.

### More styling
Added a bunch of new styles and some markup. Removed some duplicate / unneeded styles. 

I added all new styles to the scss files, and avoided any styling inside the .js files – haven't quite wrapped my head around when and how to use that. Help appreciated!


bin/watch-css
```diff
@@ -0,0 +1 @@
+nodemon -e scss -x "npm run build-css"
```
juliangruber: we can move those two scripts into the package.json, so there's less files

lib/header.js
```diff
@@ -24,13 +23,22 @@ module.exports = (props) => {
...
   };
   return yo`
-    <header style=${css(style)}>
-      <h1>All Dats</h1>
-      ${button({
-        text: 'Create new Dat',
-        click: props.create
-      })}
-      <input type="text" placeholder="Import dat" onkeydown=${keydown}>
+    <header style=${css(style)} class="dat-header">
```
juliangruber: i'll remove this inline styles logic after we merge this. the idea was to replicate what react folks are doing (see https://speakerdeck.com/vjeux/react-css-in-js). ideally we'll eventually come up with a way to merge both appraoches, giving you the flexibility of sass _and_ all the other merits from the css-in-js idea

public/css/main.css
```diff
@@ -0,0 +1,883 @@
+/*! normalize.css v4.1.1 | MIT License | github.com/necolas/normalize.css */
```
juliangruber: can we not check in built css? imo files created by the build process make diffs harder to read and aren't necessary to be checked in. if you agree, we can just add this file to the `.gitignore`

scss/hyperdrive-ui.scss
```diff
@@ -0,0 +1,147 @@
+#hyperdrive-ui {
```
juliangruber: why did you choose `hyperdrive-ui` as a class name?

juliangruber:
awesome! i added a couple of notes and questions, but otherwise this looks very very good to me :) great work!

lib/header.js
Kriesse: Sounds good! Yeah would love to discuss how to get the best out of both worlds.

public/css/main.css
Kriesse: ugh, yeah, sorry forgot to take care of that!

scss/hyperdrive-ui.scss
Kriesse: This was the css that (I believe) was originally copied over from hyperdrive-ui. The wrapping class was a quick fix to avoid any of the styles inside interfere with the new dat-desktop styles (e.g. the global styles for table elements).
I didn't want to delete it yet because I wasn't sure whether you had planned to render any parts of hyperdrive that will need those styles, so using that id would've enabled these styles.

tl;dr: This file can be deleted entirely if you don't need the styles in there :)

Kriesse:
Yay, thanks for checking this out and the detail feedback, replied to the comments!

juliangruber:
ok, can you clean up the parts we discussed? so the commit history is cleaner.

i think in the future it's easier if you create your branches based on this repo, so i can push modifications to them. but for now, would you do the honor? :)

scss/hyperdrive-ui.scss
juliangruber: yeah i think we can get rid of this then, let's add it back if we need it

Kriesse:
Of course! Done.

juliangruber:
🎆

PR #31, opened 2016-07-15
Styles

Kriesse:
Not sure any of this still makes sense with the button and import-dat input moved into dat-header. 
Might want to just keep this one: https://github.com/juliangruber/dat-desktop/commit/5b297d491c13790d8236749b574ecd99305fefc9


karissa:
i think it's smart to move them into smaller/different components since you said they'll diverge from eachother

juliangruber:
what, i did not add that in progress label oO

juliangruber:
IMO this still makes sense, this is the app, this has the app specific styles

juliangruber:
@Kriesse what's the current state of this, with all the changes in dat.land?

Kriesse:
@juliangruber Yeah let's ditch this one.

PR #32, opened 2016-07-18
Revert "refactor out header.js into dat-header"

Kriesse:
As per an earlier discussion with @juliangruber:

I suggest to revert this (at least for now). Sharing a header module between dat-desktop and dat.land seems to complicate things, as the headers will most likely have a different layout and maybe even consist of different components.

As currently things are still moving a lot and we are in exploration mode, I suggest to stick to this workflow for styles, which makes it much easier for me to iterate quickly:

- Keep all styles that are reusable in dat-design. Use classes, keeping in mind that they should be reusable and component-oriented (I follow this flavor of the [BEM naming convention](http://csswizardry.com/2013/01/mindbemding-getting-your-head-round-bem-syntax/), which is not pretty but helps to keep our CSS more robust and readable as long as we have it in one global file).
- Make use of those styles in dat-desktop and dat.land by attaching CSS classes to the markup. 
- In dat-desktop/dat.land, add or overwrite any styles that are relevant to the app only, and won't be reused elsewhere.

### DIFF
--- a/lib/header.js
+++ b/lib/header.js
@@ -1,0 +1,44 @@
+'use strict';
+
+const yo = require('yo-yo');
+const button = require('dat-button');
+const css = require('yo-css');
+const encoding = require('dat-encoding');
+
+const style = {
+
+};
+
+module.exports = (props) => {
+  const keydown = (e) => {
+    if (e.keyCode === 13) {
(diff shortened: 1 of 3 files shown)


karissa:
+1

PR #45, opened 2016-09-09
use hex strings to be compatible with browser and cli

karissa:
but it makes the sharing box ugly..

screen shot 2016-09-09 at 4 58 32 pm

### DIFF
--- a/app.js
+++ b/app.js
@@ -12,8 +12,8 @@ const render = require('./lib/render')
 const minimist = require('minimist')
 const exec = require('child_process').exec
 const raf = require('random-access-file')
-const swarm = require('hyperdrive-archive-swarm')
 const encoding = require('dat-encoding')
+const swarm = require('hyperdrive-archive-swarm')
 const hyperImport = require('hyperdrive-import-files')
 const rmrf = require('rimraf')
 const assert = require('assert')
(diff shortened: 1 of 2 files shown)


juliangruber:
Is the problem here that dat-desktop uses the new encoding, or that browser and cli use the old one?

I'm down with using hex encoding for now, to ease the transition period, but I'm getting the feeling that the new encoding might not really happen, because of friction?

I'll look into styling this a bit better, maybe just putting it into a text box.

karissa:
the cli uses hex strings, so i think we should tailor the desktop/browser to be the same @mafintosh

karissa:
+1 on text box!

karissa:
As far as internal/external, it makes sense to use the hex strings everywhere. 

I was under the impression that the encoding is useful for doing subdomains, but we decided to use hex strings as the defacto standard for now.

juliangruber:
> but we decided to use hex strings as the defacto standard for now.

ah gotcha, I didn't get that. I'll deprecate dat-encoding then and go for hex strings all the way. Or do you think we should keep using a module for this, so it's easier to change things when necessary?

karissa:
It would be nice if there was a module, but we do .toString('hex') everywhere including in the cli so it would be a bit of a project to migrate. Should at some point, though IMO.. :)

juliangruber:
https://github.com/juliangruber/dat-encoding/pull/10

karissa:
great

karissa:
can we close this if you published dat-encoding?

juliangruber:
not yet, i want to try to make this a non-breaking change

juliangruber:
it did become a breaking change, but hopefully the last. i'm releasing now.

juliangruber:
i have the release ready, just need to wait until i have faster wifi again. uploading currently times out.

PR #55, opened 2016-10-09
auth0 integration

juliangruber:
This is a WIP pull request.

The login fails currently although my details are correct, maybe there's something wrong with our api endpoint? You can just run this and it will immediately prompt you for credentials, so you can give it a try, @karissa


karissa:
seems to be working for me, what were you trying to do? Did you try creating an account first?

karissa:
Ah ok It seems that the bug is the popup is going blank the first time, then if you cancel and try again, it goes through the second time... no errors.

karissa:
found the bug cross referenced here: https://github.com/auth0/auth0-electron/issues/16

karissa:
looks like it works now with `sso: false`

karissa:
ok updated the theme, too. 
screen shot 2016-10-17 at 6 57 49 pm

clkao:
works great! we should probably have logout and a way to configure different client key based on dev/prod before this gets merged.

karissa:
ok, added simple logout button but it needs to be actually designed. going to wait on choo for that.

I also added dev/prod environment stuff.

in the meantime i think we can merge.

PR #62, opened 2016-10-21
fix(views): rename lib to views in code

yoshuawuyts:
267deb2 introduced a regression by renaming directories while code
wasn't renamed. This patch fixes that. Thanks!


karissa:
Oops nice catch :)

PR #71, opened 2016-10-25
Choo choo choo

yoshuawuyts:
very much wip, creating this PR so I can track the changes made
### Changes
- use `choo@4`
- ~~serve browserify bundle over http on port `8093` so `electron` becomes closer to a regular browser env - we can optimize this later for production (e.g. prebundle and the likes)~~
- switched dir structure a little bit - `views` is now `elements` to put in line with `dat-elements`, added a `pages` as a 1:1 reference to the pages our router renders
- nope actually let's not use browserify quite yet, see #72


yoshuawuyts:
ok, like 70% there I think - left the cafe while running into a bug. Gonna make some food now, and then get back to this. Todo:
- fix that bug (something with args passing)
- move `_app.js` into `app.js` (this is a big-ish one)
- split out RPC architecture into separate models and use event architecture
- validate deps and remove what we don't need anymore (oops, browserify)

At the end we'll have the current app using `choo`. Couple more hours of focused work is my guess based on how long it's taken to get to this point

yoshuawuyts:
Alright, I think this is ready for review now. All functionality should work as expected. It might be a bit rough around the edges in terms of performance, but given we're moving the logic out to to `dat-js` I spent no time optimizing (perceived) performance, and did a reasonably raw port of the prior code.

The commits can be rebased using the rebase merge feature. We should probably not release this version as it has no gains for end users (and perf regressions might occur) - I think the round after this might be a good candidate for releasing tho.

Keen to hear what humans think of the code! :sparkles:

cc/ @juliangruber @karissa @Kriesse

lib/create-archive.js
```diff
@@ -0,0 +1,26 @@
...
+
+// create a new archive
+// (obj, obj{str, str, fn}) -> obj
+function createArchive (drive, { path, key, isFile }) {
+  assert.ok(drive, 'lib/create-archive: drive is not defined')
+  assert.ok(path, 'lib/create-archive: path is not defined')
+  // TODO(yw): figure out why a key's not always passed - probs irrelevant
+  // once we start using dat-js
+  // assert.ok(key, 'lib/create-archive: key is not defined')
+  assert.ok(isFile, 'lib/create-archive: isFile is not defined')
```
karissa: if isFile is false, this always throws. So if you share a directory, it will crash. isFile is a boolean. It probably should check `isFile !== undefined` or something like that.

package.json
```diff
@@ -6,38 +6,49 @@
...
     "rebuild": "npm rebuild --runtime=electron --target=1.1.3 --disturl=https://atom.io/download/atom-shell --abi=48 && cd node_modules/appdmg/ && rm -Rf node_modules && npm install",
-    "build-css": "node-sass --importer node_modules/node-sass-magic-importer --include-path scss scss/main.scss public/css/main.css",
+    "start": "NODE_ENV=development electron .",
+    "test": "standard && npm run deps",
     "watch-css": "nodemon -e scss -x \"npm run build-css\""
   },
   "devDependencies": {
     "appdmg": "^0.4.5",
+    "bankai": "^3.3.1",
+    "browserify": "^13.1.1",
```
karissa: lets remove the browserify/bankai dependencies if we aren't using them

karissa:
LGTM!

lib/create-archive.js
karissa: Ah, but isFile can be undefined if you use the 'import dat' box

yoshuawuyts:
feedback incorporated

yoshuawuyts:
Ok, merging! :sparkles:

yoshuawuyts:
@karissa could you perhaps add me / the team to the repo? Couldn't merge it myself. Thanks! :grin:

juliangruber:
(tried to add yosh but also don't have admin rights)

PR #73, opened 2016-11-02
port to dat-js

juliangruber:
This is a WIP pull request for the dat-js migration, as mentioned in  https://github.com/datproject/dat-desktop/issues/44. A lot of things work already, but there's a bunch of TODOs left.

TODO:

- [x] Rendering is slow, with 3 dats loaded and ~10 calls to the render function, >120s are spent inside the deep diffing library. @yoshuawuyts i think the reducing i'm doing [here](https://github.com/datproject/dat-desktop/pull/73/files#diff-4118905b777c02ec9a054cf17443d8d0R33) isn't integrated properly and should be happening somewhere else?
- [x] archives aren't downloaded yet, I don't know why. Maybe calling `archive.share()` and `archive.download()` simultaneously isn't supported?
- [x] <del> add single file support to dat-js: https://github.com/joehand/dat-js/issues/29</del>
- [x] allow buffer .key in dat-js: https://github.com/datproject/dat-js/pull/36

Wait for API refactor:

- [ ] add live streams to dat-js: https://github.com/joehand/dat-js/issues/30
- [ ] close db in dat-js: https://github.com/joehand/dat-js/issues/18
- [x] expose archive key in dat-js: https://github.com/joehand/dat-js/issues/33
- [x] expose archive owner in dat-js: https://github.com/joehand/dat-js/issues/34
- [ ] expose .downloaded in dat-js: https://github.com/datproject/dat-js/pull/39


yoshuawuyts:
@juliangruber nice work kicking it off and summarizing it this way :tada:

yoshuawuyts:
I'm not sure if this is something that should be picked up with this patch; but the way I got `choo` re-renders to work is by incrementing [a counter](https://github.com/datproject/dat-desktop/blob/master/models/app.js#L27) every time `lib/live-stream.js` [would call `cb()`](https://github.com/datproject/dat-desktop/blob/master/lib/live-stream.js#L32).

This is because that part of the code would run mutations of `state` and we needed to trigger a re-render. What might be nice is if we could make it so that our interface layer with `dat-js` can make it so that we can state mutations back into the `choo` layer using `reducers` rather than relying on counters to trigger re-renders; feel it would end up making for easier to reason code :sparkles:

Hope I make sense haha; heaps good work so far!

pages/main.js
```diff
@@ -53,22 +53,21 @@ function createTable (dats, send) {
       <tr class="dat-list__item">
         <td>
           <div class="dat-hexagon">
-            ${dat.downloaded ? '↑' : '↓'}
+            ${'TODO: always false' && dat.archive && dat.archive.downloaded ? '↑' : '↓'}
```
maxogden: fwiw i use git blame for o/

juliangruber: same, just git blame for me

yoshuawuyts: cooool :tada:

juliangruber: ideally those won't be merged into master anyway :)

joehand:
>  (blocking) archives aren't downloaded yet, I don't know why. Maybe calling archive.share() and archive.download() simultaneously isn't supported?

This is not supported. I can't quite tell what you are trying to do. `archive.share()` will try to add files. `archive.download()` will connect to swarm and download updates.

joehand:
>  (blocking) add single file support to dat-js: datproject/dat-js#29

Where do you store the hyperdrive db for this right now? 

I implemented this before in CLI but because it was so inconsistent with the `.dat` folder it didn't make much sense and I removed it.

karissa:
Yes I'd agree that the single file support is not a blocker and should be removed from the feature list.

If you create a dat it's a share, if you import one it's a download. Once we have multi writer the new dat JS API will make more sense and we will update it to do both share and download at the same time.

juliangruber:
> This is not supported. I can't quite tell what you are trying to do. archive.share() will try to add files. archive.download() will connect to swarm and download updates.

Previously dat-desktop was architected in a way that no matter if you create or download an archive, it will be added to the internal db and then downloaded _and_ shared. One of those will be a noop, like if you share an archive, downloading won't do anything, but that didn't matter much.

I'll just change it to share if `archive.owner`, otherwise download.

juliangruber:
open for review again :)

lib/live-stream.js
```diff
@@ -31,34 +29,57 @@ function streamData (args, archives, cb) {
...
-            cb()
+        const archive = args.createArchive(Object.assign({ key }, data.value))
+        archive.open(err => {
+          if (err) throw err // TODO
+
+          // TODO add to dat-js https://github.com/joehand/dat-js/issues/30
+          archive.listStream = archive.archive.list({ live: true })
+          archive.listStream.on('data', entry => {
+            if (entry.name !== 'dat.json') return
+            collect(archive.archive.createFileReadStream('dat.json'), (err, raw) => {
```
karissa: archive.archive is a bit strange semantically, maybe dat.archive?

lib/live-stream.js
```diff
@@ -31,34 +29,57 @@ function streamData (args, archives, cb) {
...
-        archive.on('content', () => cb())
-        archive.listStream = archive.list({ live: true })
-        archive.listStream.on('data', entry => {
-          if (entry.name !== 'dat.json') return
-          collect(archive.createFileReadStream('dat.json'), (err, raw) => {
-            if (err) return
-            const json = JSON.parse(raw.toString())
-            archive.title = json.title
-            cb()
+        const archive = args.createArchive(Object.assign({ key }, data.value))
```
karissa: whats `Object.assign` doing here?

lib/live-stream.js
```diff
@@ -31,34 +29,57 @@ function streamData (args, archives, cb) {
...
       dat.listStream.destroy()
-      if (dat.path.indexOf(root) > -1) {
-        rmrf(dat.path, err => {
-          if (err) throw err
+      dat.close(function (err) {
+        if (err) console.error(err)
+        dat.db.close(function (err) {
+          if (err) console.error(err)
+          if (dat.path.indexOf(root) > -1) {
+            rmrf(dat.path, err => {
```
karissa: Should we remove people's files on their computer now while we're still in testing/beta mode? I think we might get some nasty complaints if people accidentally delete something important when they thought they were just removing it from the list.

lib/live-stream.js
yoshuawuyts: agree that this is a bit hard to read; perhaps using a bit less ES6 could help with this?

yoshuawuyts:
If we're porting to `dat-js` most of the `lib/` files should be removed I think. E.g. `create-archive` and `live-stream` should be replaceable I reckon. If they're not, what does it take to make them replaceable?

lib/live-stream.js
juliangruber: Effectively this is:

```js
var opts = clone(data.value)
opts.key = key
return opts
```

here.

`Object.assign()` is quite powerful, you can use it for merging and cloning objects. Since this is a native JS api and being used exactly for what it's intended for, I think this is ok. People will get familiar to it!

See https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Object/assign.

juliangruber:
> If we're porting to dat-js most of the lib/ files should be removed I think. E.g. create-archive and live-stream should be replaceable I reckon. If they're not, what does it take to make them replaceable?

Great idea!

- lib/create-archive.js:
  - expose `.path` in `dat-js`
  - persist `.owner?` flag in `dat-js`
- lib/live-stream.js:
  - Could be a `dat-js-manager` module?

karissa:
I don't think we need to persist the owner flag right now, do we? 

I'd see if there's another use case before we split out `dat-js-manager` into a separate module. 

What do we need to get this merged now? Seems to be a lot of merge conflicts

juliangruber:
> I don't think we need to persist the owner flag right now, do we?

unless we persist it, there's no easy way to tell

> What do we need to get this merged now? Seems to be a lot of merge conflicts

if we +1 this i'll merge it and get the conflicts sorted

yoshuawuyts:
yeah think it's cool to merge this (:

yoshuawuyts:
Wooh!

PR #76, opened 2016-11-03
2 package.json structure. closes #75

### LINKED ISSUE #75: two package.json structure
for https://github.com/datproject/dat-desktop/issues/42

> Why?
>
> 1. Native npm modules (those written in C, not JavaScript) need to be compiled and here we have two different compilation targets for them. Those used within the application need to be compiled against the electron runtime and all devDependencies need to be compiled against your local node.js environment. Thanks to the two package.json structure, this is trivial (see #39).
> 2. No need to specify which files to include in the app (because development files reside outside the app directory).

juliangruber:
Move to a 2 package.json structure, which is recommended for using electron-builder.

> Why?
>
> 1. Native npm modules (those written in C, not JavaScript) need to be compiled and here we have two different compilation targets for them. Those used within the application need to be compiled against the electron runtime and all devDependencies need to be compiled against your local node.js environment. Thanks to the two package.json structure, this is trivial (see #39).
> 2. No need to specify which files to include in the app (because development files reside outside the app directory).

Temporarily removed the current build scripts, which will be replaced anyway. The only things that effectively changed are:

- the source is now in `app/`
- you need to run `(cd app/ && npm install)` too


yoshuawuyts:
Don't think I agree, will review later (about to board)

karissa:
Hey, saw this one got pushed through quickly -- now that we have more hands on this, let's try to make sure that everything has been PR reviewed (essentially no direct commits, because when you open a PR and merge it quickly its almost the same as committing directly).

PR #82, opened 2016-11-17
add real progress stats

### LINKED ISSUE #81: add hyperdrive-stats
for progress information, see https://github.com/datproject/dat-desktop/issues/19#issuecomment-255356234

juliangruber:
closes #81 

Add real progress stats to the UI!

@joehand do we want to integrate this into dat-js? Would be easier to consume this way.


lib/live-stream.js
```diff
@@ -53,6 +55,11 @@ function streamData (args, archives, cb) {
               cb()
             })
           })
+          dat.hyperStats = Stats({
+            archive: dat.archive,
+            db: sub(args.db, `${encoding.encode(dat.key)}-stats`)
```
joehand: Whats the reason for using a sub level here instead of the dat.db?

joehand:
> @joehand do we want to integrate this into dat-js? Would be easier to consume this way.

Yes! I do have block progress in there now, but am planning on moving to using hyperdrive stats as one of the next things I do. 

I need to review hyperdrive stats a bit, last time I looked there was something missing but I think it's all there now. Will go through this today!

lib/live-stream.js
juliangruber: so we're sure never to have a key collision

lib/live-stream.js
```diff
@@ -53,6 +55,11 @@ function streamData (args, archives, cb) {
...
+          })
+          dat.hyperStats.on('update', () => cb())
```
yoshuawuyts: Perhaps out of scope for this patch, but this pattern feels like a smell - now that we've got dat-js in here we should think of a way to actually move data around (:

juliangruber: > we should think of a way to actually move data around (:

can you expand on that?

juliangruber:
let's iterate on this in future work

yoshuawuyts:
the cb() statement increments a counter rn to force a refresh instead of
passing new data - feel that's not ideal; let's talk about it next week
when you're around and maybs we can find something that works well (:

juliangruber:
yeah definitely! i was curious about this, why do we not follow the "normal" choo pattern and just pass data around?

PR #84, opened 2016-11-18
improve progress bar and hexagon

juliangruber:
cc @Kriesse

### DIFF
--- a/pages/main.js
+++ b/pages/main.js
@@ -60,16 +60,15 @@ function createTable (dats, send) {
         ? 'loading'
         : 'paused'
       : 'complete'
+    const hexContent = {
+      loading: '↓',
+      paused: '',
+      complete: '↑'
+    }[state]
     return html`
       <tr class="dat-list__item">
         <td>
-          <div class="dat-hexagon">
-            ${{
-              loading: '↓',
-              paused: '',
-              complete: '↑'
-            }[state]}
-          </div>
+          <div class="dat-hexagon">${hexContent}</div>
         </td>
         <td>
           <div class="cell-truncate">
--- a/pages/main.js
+++ b/pages/main.js
@@ -80,7 +80,16 @@ function createTable (dats, send) {
               ${progress}%
             </div>
             <div class="progress__bar">
-              <div class="progress__line progress__line--progress" style="width: ${progress}%"></div>
+              <div
+                class="progress__line progress__line--${
+                  progress < 1
+                    ? dat.stats.peers > 0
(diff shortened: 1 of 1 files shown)


pages/main.js
```diff
@@ -55,11 +55,20 @@ function createTable (dats, send) {
...
     return html`
       <tr class="dat-list__item">
         <td>
           <div class="dat-hexagon">
-            ${dat.owner ? '↑' : '↓'}
+            ${{
+              loading: '↓',
+              paused: '',
+              complete: '↑'
+            }[state]}
```
yoshuawuyts: I find this hard to read

juliangruber: it's something I learned from @substack years ago, once you get used to it it's a lot easier to read and type than say

```js
${
  state === 'loading' && 'v' ||
  state === 'paused' && '' ||
  state === 'complete' && '^'
}
```

it's basically a switch statement as a value, which JavaScript is missing.

karissa: Works for me

yoshuawuyts:
hah yeah I get what you're doing and it's clever - but clever code is also
hard to debug. What I'd propose is that instead of selecting the value
inline you'd define it as a variable somewhere higher in the function and
use that

clkao:
hey @juliangruber the dependency-ci complains that `hyperdrive-stats` is not licensed

juliangruber:
> hey @juliangruber the dependency-ci complains that hyperdrive-stats is not licensed

thanks, fixed!

juliangruber:
> hah yeah I get what you're doing and it's clever - but clever code is also
> hard to debug. What I'd propose is that instead of selecting the value
> inline you'd define it as a variable somewhere higher in the function and
> use that

@yoshuawuyts you mean like this?

```js
const hexContent = {
  loading: '↓',
  paused: '',
  complete: '↑'
}[state]
```

yoshuawuyts:
Yeah, doing that inline is hard to read. Meant to reply to the code snippet
but apparently email doesn't let me hah

juliangruber:
cool, check this out @yoshuawuyts

karissa:
Yeah +1, I'm down with pulling all js out of the templates if possible, so the html can focus on being html while variables are defined above. Thanks for pointing that out, @yoshuawuyts !

clkao:
LGTM. The progress is not actually updating (numbers in stats object stays the same), but this is probably another issue.

juliangruber:
@clkao it is, just taking a long time. I haven't gotten behind that yet.

PR #88, opened 2016-11-21
feat(index): use electron-window

yoshuawuyts:
Lil maintenance patch to clean up `index.js` - gives us some room to add more features. Think this is cool because it turns the entry file for electron into something that weighs 30 lines of code that's easy to follow :sparkles:

## Changes
- use `envobj` to validate environment variables
- create `lib/delegate-electron-events` to wrap boilerplate in `index.js`, needed to safely delegate file and link events sent to electron while booting (should turn into a package at some point probably)
- use `electron-window` to remove some of the browser boot boilerplate

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,60 +1,30 @@
-const {app, BrowserWindow, ipcMain, Menu, shell} = require('electron')
 const defaultMenu = require('electron-default-menu')
+const { app, shell, Menu } = require('electron')
+const window = require('electron-window')
+const Env = require('envobj')
+const path = require('path')
 
-const env = process.env.NODE_ENV
+const delegateEvents = require('./lib/delegate-electron-events')
 
-let win, file, link
-
-// handle electron events
-ipcMain.on('ready', () => {
-  if (file) {
-    let path = file
-    file = null
-    win.webContents.send('file', path)
-  }
-  if (link) {
-    let url = link
-    link = null
-    win.webContents.send('link', url)
-  }
-})
-
-// create the HTML window inside the electron client
-// null -> null
-function createWindow () {
-  win = new BrowserWindow({ width: 1000, titleBarStyle: 'hidden-inset' })
-
-  // TODO(yw): move the HTML file to a separate package
-  win.loadURL(`file://${__dirname}/index.html`)
-  if (env === 'development') {
-    win.webContents.openDevTools({
-      mode: 'detach'
-    })
-  }
-  win.on('closed', () => { win = null })
-
-  Menu.setApplicationMenu(Menu.buildFromTemplate(defaultMenu(app, shell)))
+const windowStyles = {
+  width: 1000,
+  titleBarStyle: 'hidden-inset'
 }
 
+const env = Env({ NODE_ENV: 'production' })
+const emitter = delegateEvents() // make sure we don't miss events while booting
+
 app.on('ready', () => {
-  app.ready = true
-  createWindow()
-})
-app.on('window-all-closed', () => {
-  if (process.platform !== 'darwin') app.quit()
-})
-app.on('activate', () => {
-  if (!win && app.ready) createWindow()
-})
-app.on('will-finish-launching', () => {
-  app.on('open-file', (ev, path) => {
-    ev.preventDefault()
-    if (win) win.webContents.send('file', path)
-    else file = path
-  })
-  app.on('open-url', (ev, url) => {
-    ev.preventDefault()
-    if (win) win.webContents.send('link', url)
-    else link = url
+  const mainWindow = window.createWindow(windowStyles)
+  const indexPath = path.join(__dirname, 'index.html')
+
+  emitter.on('open-file', (file) => mainWindow.webContents.send('file', file))
+  emitter.on('open-url', (url) => mainWindow.webContents.send('link', url))
+
+  mainWindow.showUrl(indexPath, () => {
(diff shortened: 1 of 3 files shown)


yoshuawuyts:
oh just now realized `ipcMain.on('ready')` is a non-standard thing. @juliangruber any thoughts on this patch?

index.js
```diff
@@ -1,60 +1,29 @@
-const {app, BrowserWindow, ipcMain, Menu, shell} = require('electron')
 const defaultMenu = require('electron-default-menu')
+const window = require('electron-window')
+const electron = require('electron')
+const Env = require('envobj')
```
juliangruber: I don't think we need this module right now, it's just one environment variable?

index.js
```diff
@@ -1,60 +1,29 @@
...
+const path = require('path')
 
-const env = process.env.NODE_ENV
+const delegateEvents = require('./lib/delegate-electron-events')
 
-let win, file, link
+const shell = electron.shell
+const Menu = electron.Menu
+const app = electron.app
```
juliangruber: it's common electron style to do the destructuring import, why did you change that?

lib/delegate-electron-events.js
```diff
@@ -0,0 +1,47 @@
...
+const electron = require('electron')
+
+const ipcMain = electron.ipcMain
+const app = electron.app
+
+module.exports = delegateElectronEvents
+
+// delegate electron events while booting
+// until a window is ready that can handle them
+// null -> EventEmitter
```
juliangruber: we're not using those api comments in this project yet, either we should add them everywhere or stick to the coding style

lib/delegate-electron-events.js
```diff
@@ -0,0 +1,47 @@
+const EventEmitter = require('events').EventEmitter
+const electron = require('electron')
+
+const ipcMain = electron.ipcMain
+const app = electron.app
```
juliangruber: same with imports

index.js
```diff
@@ -1,60 +1,29 @@
...
-    win.webContents.send('link', url)
-  }
-})
-
-// create the HTML window inside the electron client
-// null -> null
-function createWindow () {
-  win = new BrowserWindow({ width: 1000, titleBarStyle: 'hidden-inset' })
+const env = Env({ NODE_ENV: 'production' })
+const emitter = delegateEvents() // make sure we don't miss events while booting
```
juliangruber: that's a cool idea

index.js
```diff
@@ -1,60 +1,29 @@
...
-  // TODO(yw): move the HTML file to a separate package
-  win.loadURL(`file://${__dirname}/index.html`)
-  if (env === 'development') {
-    win.webContents.openDevTools({
-      mode: 'detach'
-    })
-  }
-  win.on('closed', () => { win = null })
+app.on('ready', () => {
+  const mainWindow = window.createWindow({ width: 1000, height: 400 })
```
juliangruber: nice!

index.js
yoshuawuyts: Figured it'd serve well as documentation also; but yeah I get your point - happy to strip it out

index.js
yoshuawuyts: Oh, I didn't consciously change it, sorry about that - I just rewrote this file from scratch and incrementally added requires as I needed them like I normally do.

That said I do dislike multiple variable declarations on a single line as I find them hard to read and modify. I believe in most of our code we aren't using this, so only using it in this particular case feels like an odd one out.

I think most people on the team (myself included) might not necessarily be familiar with how destructuring works in detail. E.g. saw this in a tweet the other day, had no idea what was going on 😕:

[cxtvcqwwqaawkl3](https://twitter.com/WickyNilliams/status/798497274653503488)

But yeah guess since you seem to feel strongly about this I'll add them back; I'd rather not get stuck arguing syntax haha :sparkles:

lib/delegate-electron-events.js
yoshuawuyts: yeah sure I'll strip 'em

index.js
karissa: it seems pretty lightweight and sets it up so its easier in the future to add environment variables. I think it's fine to keep it if we think we're going to have more env vars.

index.js
karissa: Maybe we want to come up with a style guide that we all agree on, I'd reckon we stick with standard style, and I'm not sure what the take is on this. As long as I don't have to use babel to compile anything, I'm fine using some of this ES6 stuff.

Wow that code in that tweet is horrific..

lib/delegate-electron-events.js
karissa: what api comments?

I'm actually very happy with starting to use jsdoc right now so we have some self documenting code if y'all want :)

karissa:
Thanks for doing this, woot!

index.js
clkao: I think the standard style guide doesn't cover destructuring.  I think one level for same-name destructuring is fine in general.

Hint: to work out the nested one, just image you are flipping the LHS and RHS, so it's like grabbing `inputClass = manager.options.inputClass`.  Though expressive, this is far too much mental exercise that reduces readability

index.js
clkao: Does this preserve the hidden-inset titleBarStyle?

yoshuawuyts: good point; missed that - fixed!

yoshuawuyts:
ok fixed all comments - thanks for reviewing y'all!

PR #89, opened 2016-11-22
Design cleanup

Kriesse:
Removed the dat-design dep, cleaned up messy styles, started to use tachyons.
This should help to make the transition to sheetify/base-elements easier and avoid breaking all the styles.


pages/main.js
```diff
@@ -95,27 +91,27 @@ function createTable (dats, send) {
...
             text: '',
-            klass: 'row-action',
+            cls: 'row-action',
             click: () => send('app:open', dat)
           })}
           ${button({
             icon: 'link',
             text: '',
-            klass: 'row-action',
+            cls: 'row-action',
```
yoshuawuyts: these `cls` properties can safely be named `class` which might make them a bit easier to understand

scss/imports/_header.scss
```diff
@@ -0,0 +1,23 @@
...
+
+.menu-trigger {
+  height: 2rem;
+  padding-left: .625rem;
+  width: 2.5rem;
+  .icon-menu {
+    fill: $color-neutral--20;
+  }
+  &:hover, &:focus {
+    .icon-menu {
```
yoshuawuyts: I do find these statements a bit hard to read, I think this translates to:
```css
.menu-trigger:hover .icon-menu,
.menu-trigger:focus .icon-menu {}
```
right? Just wanna make sure I understand what's going on alright haha

pages/main.js
juliangruber: doesn't that mess with anything that tries to parse the source? `class` is a reserved keyword

app/public/css/main.css
```diff
@@ -0,0 +1,1734 @@
+/**
```
juliangruber: is this file hand written or generated?

yoshuawuyts:
Nope it's fine, reserved keywords as property names were ok going ES5
onward iirc

juliangruber:
> Nope it's fine, reserved keywords as property names were ok going ES5
> onward iirc

Oh, learned something :) Ok if this is only ever going to be used in Dat Desktop, I'm +1, otherwise we do need to worry about browsers and old node versions.

yoshuawuyts:
Oh yeah afaik this runs like IE7+ or something / all node versions haha -
like ES5 has been around for a while I think. I've never run into issues at
least (:

app/public/css/main.css
Kriesse: @juliangruber it's generated – i guess the entire `app` directory should be gitignored? (where does that even come from? i'm confused)

scss/imports/_header.scss
Kriesse: Yup, that's understood correctly. I love the `&` selector for this particular use case: I almost always use the same styles for `&:hover, &:focus` for easy accessibility wins and find the grouping helpful – but can see that it’s a bit hard to read if not used to it.
Can we have `&` in sheetify at all?

app/public/css/main.css
juliangruber: oh! it's a leftover of a change that was reverted already. you can safely remove it

karissa:
OK I fixed the merge conflict and cleaned up the unneeded main.css file

PR #91, opened 2016-11-24
Remove live stream architecture

juliangruber:
Bunch of stuff here:

- Remove live stream architecture so we don't run into weird issues having to call `dat.open -> dat.close -> dat.open` and stuff like that
- Make `lib/db.js` export a db instead of a function that returns a db, so modules can just `require('./db')` instead of us having to pass around the db instance everywhere
- Same with `lib/root-dir.js`, but also `fs.mkdirSync()` the root dir on start up in this library
- Move ipc stuff to model, as @yoshuawuyts already commented
- Create a dat manager library that should eventually be its own thing

I need to put this through more testing first, but so far I think this is going in a good direction.


lib/dat-manager.js
```diff
@@ -0,0 +1,138 @@
...
+    if (Buffer.isBuffer(key)) key = encoding.encode(key)
+    const dat = Dat({ key, dir })
+    dat.dir = dir
+    dat.owner = owner
+    if (dat.owner) {
+      dat.share(err => {
+        if (err) dat.emit('error', err)
+      })
+    } else {
+      dat.download(err => {
```
karissa: What if we put these share/download calls into activate? Then load doesn't start any processes

lib/dat-manager.js
```diff
@@ -0,0 +1,138 @@
...
+      dat.hyperStats.on('update', () => dat.emit('update'))
+      dat.on('files-counted', () => dat.emit('update'))
+      dat.on('archive-finalized', () => dat.emit('update'))
+      dat.on('archive-updated', () => dat.emit('update'))
+      dat.on('download-finished', () => dat.emit('update'))
+      dat.on('swarm-update', () => dat.emit('update'))
+    })
+    return dat
+  }
+  remove (dat, cb) {
```
karissa: Remove might be different from pause, right? I might want to stop the swarm and updates, which would mean "deactivating" - what do you think about having a separate function for that and then remove calls deactivate then removes it from the list

lib/dat-manager.js
```diff
@@ -0,0 +1,138 @@
...
+          }
+          dat.title = json.title
+          dat.emit('update')
+        })
+      })
+      dat.hyperStats = Stats({
+        archive: dat.archive,
+        db: sub(db, `${encoding.encode(dat.key)}-stats`)
+      })
+      dat.hyperStats.on('update', () => dat.emit('update'))
```
karissa: Is this going to call update a bunch of times that duplicate the events below?

lib/dat-manager.js
```diff
@@ -0,0 +1,138 @@
...
+          if (err) return console.error('failed to read dat.json', err)
+          let json
+          try {
+            json = JSON.parse(raw.toString())
+          } catch (err) {
+            console.error('failed to parse dat.json', err)
+            return
+          }
+          dat.title = json.title
+          dat.emit('update')
```
karissa: This is going to update the ui with every file which might become problematic. Do we need that right now? I think we should only update when we have to, especially since we don't preview the files right now.

lib/dat-manager.js
```diff
@@ -0,0 +1,138 @@
...
+  create (dir, cb) {
+    const dat = Dat({ dir })
+    dat.owner = true
+    dat.dir = dir
+    dat.share(err => {
+      if (err) return cb(err)
+      this.persist(dat, err => {
+        if (err) return cb(err)
+        this.dats.push(this.watch(this.activate(dat)))
+        this.emit('update')
```
karissa: Same, do we need this update?

lib/db.js
```diff
@@ -1,16 +1,9 @@
...
-// str -> obj
-function createDb (rootDir) {
-  assert.equal(typeof rootDir, 'string', 'lib/db: rootDir should be a string')
-
-  return level(`${rootDir}/.db`, {
-    keyEncoding: bytewise,
-    valueEncoding: 'json'
-  })
-}
+module.exports = level(`${root}/.db`, {
```
karissa: Will this recall level every time it's required?

lib/dat-manager.js
```diff
@@ -0,0 +1,138 @@
...
+      lt: ['archive', undefined]
+    })
+    .on('data', data => {
+      dats.push(this.activate(this.load(data.value)))
+    })
+    .on('error', err => cb(err))
+    .on('end', () => cb(null, dats))
+  }
+  persist ({ key, owner, dir }, cb) {
+    db.put(['archive', key], {
```
karissa: Probably should check here if the key they give is the right kind of key before putting it in the db

karissa:
This looks great! I remember now that i've used ipc in an electron app to send messages a manager-like object, and it worked pretty well. Do you plan on moving the manager to the backend process?

lib/dat-manager.js
juliangruber: This causes a re render after an archive has been created. Otherwise it would only show up in the list once an internal hyper* event causes an update.

lib/dat-manager.js
juliangruber: This function is only called internally, so we control the arguments that are being passed. Should probably make this more clear by using Symbols.

juliangruber: renamed to `_persist`, together with the other private functions

lib/dat-manager.js
juliangruber: This will update the UI only when the `dat.json` updates. We could check if `json.title !== dat.title` too, but eventually that will be solved by choo anyway.

lib/dat-manager.js
juliangruber: not necessarily, hyperStats is somewhat independent of those events as it works up the archive's whole history instead of only watching for the "now". We will get duplicate update events still, but that as well should be solved once choo can do proper diffing as we should switch to immutable data structures underneath.

lib/dat-manager.js
juliangruber: We don't have a pause function in the UI yet, let's add that once the product aligns?

juliangruber:
Thanks a lot for your thorough review, @karissa! I addressed all your points in code and comments :)

PR #93, opened 2016-11-30
Add browserify

juliangruber:
cc @yoshuawuyts

### DIFF
--- a/lib/transform.js
+++ b/lib/transform.js
@@ -1,0 +1,19 @@
+'use strict'
+
+const through = require('through2')
+
+module.exports = file => {
+  if (!/\.js$/.test(file)) return through()
+  let src = ''
+  const out = through((chunk, _, cb) => {
+    src += chunk
+    cb()
+  }, function (cb) {
+    src = src.replace(/require\('bindings'\)\('(.*)'\)/, (_, bindings) =>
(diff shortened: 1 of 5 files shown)


lib/transform.js
```diff
@@ -0,0 +1,19 @@
+'use strict'
```
juliangruber: this is required for the bindings module to work with browserify. Should go into a module and use the AST if we decide we want this approach.

yoshuawuyts: I'm happy to split this into a package if you want - what do you reckon would be a good name?

mafintosh: bindingify! (no idea what this does though lol)

juliangruber:
> (no idea what this does though lol)

The bindings module needs to know the `__dirname` of the module calling it. It currently uses Error stack traces to get that, but it's way cleaner to just pass the `__dirname` as an argument instead. And, that works with browserify.

juliangruber:
Simply enabling source maps doesn't do that trick either.

juliangruber:
Cool, @yoshuawuyts, are you taking over on the module?

yoshuawuyts:
@juliangruber yup, happy to do it - I propose we merge this as-is and then replace it with an external package in a separate PR :sparkles:

yoshuawuyts:
alright merging :sparkles:

PR #95, opened 2016-12-01
chore(models): use choo-model

yoshuawuyts:
Use a package rather than our hand-rolled solution


juliangruber:
..and nice clean up work!

PR #96, opened 2016-12-01
feat(deps): use bindingify

yoshuawuyts:
Use https://github.com/yoshuawuyts/bindingify rather than something inline :tada:


juliangruber:
thanks!

PR #99, opened 2016-12-05
feat(modal): split off

yoshuawuyts:
Phoo, this required a good amount of fixes to fringe libraries but think it should be all good now; replaces the modal element we were using with a new one. Would be neat if someone could try and run it to see if it works for them. Didn't really invest much into styling stuff but I figured @Kriesse would be happy to polish it to a better state haha :sparkles:

All the stuff in `lib/` should be moved into a separate package (base-elements ?), but gonna reserve that for another PR hah. Thanks!

--- 
Todo: add screenshot

### DIFF
--- a/elements/modal.js
+++ b/elements/modal.js
@@ -1,0 +1,68 @@
+const widget = require('cache-element/widget')
+const Modal = require('../lib/modal-element')
+const html = require('choo/html')
+
+module.exports = createWidget
+
+function createWidget () {
+  let link = null
+
+  let modal = Modal(html`<div></div>`, () => {
+    link = null
+  })
+
+  return widget({
+    onload: function () {
+      if (link) modal.show(createHtml(link))
+    },
+    onunload: function () {
+      link = null
+      modal.hide()
+    },
+    onupdate: function (el, newLink) {
+      if (!link && newLink) {
+        // fresh link
+        link = newLink
+        modal.show(createHtml(link))
+      } else if (!newLink) {
+        // link was cleared
+        link = null
+        modal.hide()
+      } else if (newLink && (link !== newLink)) {
+        // ohey, we moved to a different link
+        link = newLink
+        modal.show(createHtml(link))
+      }
+    },
+    render: function (newLink) {
+      link = newLink
+      if (link) {
+        modal.show(createHtml(link))
+      } else {
+        modal.hide()
+      }
+
+      return html`<div>${modal}</div>` // this fixes a bug in modal-element
+    }
+  })
+
+  function createHtml (link) {
+    return html`
+      <section class="flex flex-column items-center justify-center pa3 b--solid bg-white">
+        <h3>Your dat link:</h3>
+        <input type="text" value=${link}/>
+        <p>
+          This link has also been copied to the clipboard for your
+          convenience.
+        </p>
+        <button onclick=${handleExit} class="pointer">Ok cool</button>
+      </section>
+    `
+
+    function handleExit () {
+      link = null
+      modal.hide()
+      window.history.back()
+    }
+  }
+}
--- a/models/app.js
+++ b/models/app.js
@@ -3,9 +3,8 @@ const clipboard = require('electron').clipboard
 const ipc = require('electron').ipcRenderer
 const exec = require('child_process').exec
 const encoding = require('dat-encoding')
-const jsAlert = require('js-alert')
 const Model = require('choo-model')
-const html = require('choo/html')
+
 const Manager = require('../lib/dat-manager')
 
 module.exports = createModel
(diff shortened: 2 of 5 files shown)


Kriesse:
@yoshuawuyts Started to style the modal (loving the new sheetify flow <3), but the funky rendering issue is back.

Steps to reproduce:
* Open modal
* hover / focus around on clickable elements
This should create weird squares in the background. 
Not sure yet what is going on.

image

elements/modal.js
```diff
@@ -0,0 +1,68 @@
+const widget = require('cache-element/widget')
+const Modal = require('../lib/modal-element')
+const html = require('choo/html')
+
+module.exports = createWidget
+
+function createWidget () {
+  var link = null
+
+  var modal = Modal(html`<div></div>`, function () {
```
juliangruber: could use some es6 arrow fns here :)

elements/modal.js
```diff
@@ -0,0 +1,68 @@
+const widget = require('cache-element/widget')
+const Modal = require('../lib/modal-element')
+const html = require('choo/html')
+
+module.exports = createWidget
+
+function createWidget () {
+  var link = null
```
juliangruber: let instead of var

elements/modal.js
```diff
@@ -0,0 +1,68 @@
...
+
+  return widget({
+    onload: function () {
+      if (link) modal.show(createHtml(link))
+    },
+    onunload: function () {
+      link = null
+      modal.hide()
+    },
+    onupdate: function (el, linkArg) {
```
juliangruber: i think this would be cleaner, and not even need the comments, with a different name for `linkArg`, like `newLink`

elements/modal.js
```diff
@@ -0,0 +1,68 @@
...
+    },
+    render: function (linkArg) {
+      link = linkArg
+      if (link) {
+        modal.show(createHtml(link))
+      } else {
+        modal.hide()
+      }
+
+      return html`<div>${modal}</div>` // this fixes a bug in modal-element
```
juliangruber: can you link to the corresponding issue in modal-element here, for reference?

elements/modal.js
```diff
@@ -0,0 +1,68 @@
...
+          convenience.
+        </p>
+        <button onclick=${handleExit} class="pointer">Ok cool</button>
+      </section>
+    `
+
+    function handleExit () {
+      link = null
+      modal.hide()
+      window.history.back()
```
juliangruber: it's going back in history here, but where does it go forward?

models/app.js
```diff
@@ -49,27 +48,20 @@ function createModel (cb) {
     const dir = files[0]
     manager.create(dir, done)
   })
+
   model.effect('delete', (state, data, send, done) => {
     const dat = data
     manager.remove(dat, done)
   })
+
+  // copy a dat share link to clipboard and open a modal
```
juliangruber: if possible i'd go with as few comments as absolutely possible. shall we rename the effect to something like `share-modal`?

Kriesse:
@yoshuawuyts @juliangruber 
UX request: Can we change the modal so that it does not automatically copy the link into your clipboard? It's not clear to the user that this is going to happen when they click the link icon, so we should not take over their clipboard without confirmation. Let's have a button on the input field that copies the link, and gives a subtle notification to confirm when that happened.

Same pattern can be found when sharing the link to a googledoc.

## Mockup:

image

image

models/app.js
juliangruber: there's nothing wrong with comments per se, it's just that if code can be written in a way that comments aren't necessary, that's even better

elements/modal.js
yoshuawuyts: The issue is in `lib/` - there's no corresponding GH issue yet as I'm not sure what the cause is; you're right I should create one tho; doing that :sparkles:

models/app.js
yoshuawuyts: @juliangruber the effect has multiple behaviors; `share-` doesn't cover them all. `share-and-copy-link-to-clipboard` might be closer but that's too long. I feel a comment has added value here to describe the side effect that happens

yoshuawuyts:
@Kriesse ohhh yeah I like that - feel the UX part makes sense also. Let's add that in a PR that builds on top of this one so this doesn't float around forever :grin:

yoshuawuyts:
@juliangruber merged all your feedback; lmk if it's good now :sparkles:

yoshuawuyts:
@Kriesse also yeah I wonder what's going on on your machine; on mine it looks like:

screen shot 2016-12-06 at 22 53 38

Kriesse:
@yoshuawuyts yeah my bad, forgot to push last night – styles are up here now, still in progress: https://github.com/datproject/dat-desktop/tree/dat-link-modal

Issue #68, opened 2016-10-21
Style Empty States


Kriesse:
When a user opens Dat Desktop for the first time, what do they see?
When the list of dats is empty, what do they see?

- [x] style empty table state
- [x] style welcome screen

Kriesse:
@yoshuawuyts I'd like two different pages here.

## 1. Welcome Page
User opens app for the first time and sees a welcome message. Buttons in header are hidden. The button removes this page and goes to the Empty State. Later we could add a more detailed click through tutorial. 
![image](https://cloud.githubusercontent.com/assets/53684/20951364/6dda79b6-bc79-11e6-817f-a5d54a3a8410.png)

## 2. Empty State
When there are no dats, user sees this (instead of an empty table). 
![image](https://cloud.githubusercontent.com/assets/53684/20951382/83025160-bc79-11e6-98ea-27ce048e384c.png)

yoshuawuyts:
alright, #111 was merged - can all be styled now (:

PR #105, opened 2016-12-07
Vanilla empty state

yoshuawuyts:
- moved the table element from the main view to `elements/table.js`
- created an empty state for the main view that's shown if there are no elements in the thing - meant to be styled by @kriesse as per #68 
- use the `prettier-bytes` package to make our byte display a bit less noisy
- "delete a dat" is broken as per https://github.com/datproject/dat-desktop/pull/91 - fixed some of the symptoms but it's still not working - opened an issue for it here https://github.com/datproject/dat-desktop/issues/104

Thanks!

### DIFF
--- a/elements/table.js
+++ b/elements/table.js
@@ -1,6 +1,6 @@
 const encoding = require('dat-encoding')
+const bytes = require('prettier-bytes')
 const html = require('choo/html')
-const bytes = require('bytes')
 
 const button = require('../elements/button')
 
--- a/elements/table.js
+++ b/elements/table.js
@@ -2,7 +2,7 @@ const encoding = require('dat-encoding')
 const bytes = require('prettier-bytes')
 const html = require('choo/html')
 
-const button = require('../elements/button')
+const button = require('./button')
 
 module.exports = tableElement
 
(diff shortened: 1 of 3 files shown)


elements/table.js
```diff
@@ -0,0 +1,116 @@
+const encoding = require('dat-encoding')
+const bytes = require('prettier-bytes')
+const html = require('choo/html')
+
+const button = require('../elements/button')
```
juliangruber: can be simplified to `require('./button')`

lib/dat-manager.js
```diff
@@ -56,21 +56,15 @@ module.exports = class Manager extends EventEmitter {
     })
   }
   remove (dat, cb) {
+    const self = this
     dat.listStream.destroy()
-    dat.close(err => {
+    dat.close(function (err) {
```
juliangruber: woot?

yoshuawuyts: oh lol

elements/table.js
yoshuawuyts: good point!

yoshuawuyts:
@Kriesse should be able to style it now :D

PR #106, opened 2016-12-07
fix(dat-manager): fix removing a dat

yoshuawuyts:
Closes https://github.com/datproject/dat-desktop/issues/104 - thanks!


yoshuawuyts:
cc/ @juliangruber if you could review would be grand :grin:

PR #107, opened 2016-12-08
initial styles for empty table state #68

Kriesse:
styles for tutorial text are there too, but still hidden because not final.


pages/main.js
```diff
@@ -9,6 +10,66 @@ const Modal = require('../elements/modal')
 
 const modal = Modal()
 
+css('dat-colors')
+
+var prefix = css`
```
juliangruber: also, should be a `const`, see https://github.com/datproject/dat-desktop/pull/108#discussion_r91483407

PR #108, opened 2016-12-08
Styles for copy dat link modal

Kriesse:


elements/modal.js
```diff
@@ -1,6 +1,122 @@
 const widget = require('cache-element/widget')
 const Modal = require('../lib/modal-element')
 const html = require('choo/html')
+var css = require('sheetify')
```
juliangruber: should be a `const`. `const` is more explicit than `var` because it prevents the variable from ever being assigned to anything else.

elements/modal.js
```diff
@@ -1,6 +1,122 @@
...
+
+const prefix = css`
+  :host {
+    --input-height: 2rem;
+    --icon-height: 1.2rem;
+    --button-width: 2rem;
+    box-shadow: 0 1.2rem 2.4rem rgba(0,0,0,.5);
+    position: relative;
+  }
+  .exit {
```
yoshuawuyts: Should we prefix these with `:host` also? Could prevent possible conflicts; though not fussed about it tho hah

Kriesse: Ha yeah I was wondering the same but then noticed they get prefixed anyway, right?
See example: 
image

Issue #103, opened 2016-12-06
[ux] style copy-link modal


yoshuawuyts:
different folks can do different parts, but I think this is what needs to happen to style the modal a bit further

### todo
- [x] stop copying link by default
- [x] build out a little model for the modal
- [x] style elements
- [x] fix regression with weird visual background things happening

---

### screenshots
![](https://cloud.githubusercontent.com/assets/53684/20926558/0c59dcd0-bc11-11e6-85bc-18be18478895.png)
![](https://cloud.githubusercontent.com/assets/53684/20926562/153ee8cc-bc11-11e6-9223-ba09d64b13e4.png)

Kriesse:
@yoshuawuyts nice! styling work started here: https://github.com/datproject/dat-desktop/tree/dat-link-modal

Kriesse:
@yoshuawuyts Can you make it so that when the link is copied to the clipboard, the class `show-confirmation` is added here? https://github.com/datproject/dat-desktop/blob/master/elements/modal.js#L170

yoshuawuyts:
@Kriesse done! - airport wifi is blocking `git push` tho (no UDP connections) so will push when home :D - really liking how it's shaping up :sparkles:

<img width="1280" alt="screen shot 2016-12-08 at 17 26 52" src="https://cloud.githubusercontent.com/assets/2467194/21018022/a3965a82-bd6b-11e6-90bf-72abe5f48cbf.png">

yoshuawuyts:
@Kriesse in terms of CSS code, I noticed we got [about 100 lines of CSS inline](https://github.com/datproject/dat-desktop/blob/master/elements/modal.js#L8-L119) for the modal - would it perhaps be an idea to replace most of it with tachyons before we close this issue?

Kriesse:
@yoshuawuyts I'm still going back and forth trying to come up with a set of rules for when to use a tachyons class vs when to use a custom class. Went through the CSS and cleaned up where it made sense, also removed all scss, please review :) https://github.com/datproject/dat-desktop/pull/110

So far, it turns out certain styles that our UI requires don't come via tachyons. So we have to add them as custom rules.
Example: `border: 1px solid var(--color-neutral-30);`
In the above example, we could of course use tachyons to set `border-width` and `border-style` on the element, and only add a custom rule for `border-color`. 
Another example:
````css
position: relative;
top: -.3rem;
````
[...]

PR #109, opened 2016-12-08
feat(modal): show dialog text

yoshuawuyts:
Implements the final modal features. Ref #103

### DIFF
--- a/elements/modal.js
+++ b/elements/modal.js
@@ -1,3 +1,4 @@
+const clipboard = require('electron').clipboard
 const widget = require('cache-element/widget')
 const Modal = require('../lib/modal-element')
 const html = require('choo/html')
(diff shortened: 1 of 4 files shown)


elements/modal.js
```diff
@@ -121,15 +122,17 @@ const prefix = css`
 module.exports = createWidget
 
 function createWidget () {
+  let isCopied = false
```
juliangruber: i wonder if this is a little over engineered, couldn't the flow also just be `open modal -> not copied. copy -> copied. close modal -> reset state`

yoshuawuyts: That's what we're doing right? Or are you talking about the unidirectional rendering stuff being too complicated? - I think I definitely prefer explicit state and re-rendering over using a selector and some sort of "toggle class" method.

Would perhaps adding a method called `reset` and calling that when we reset stuff help?

yoshuawuyts:
Did a round of updates incorporating @juliangruber's feedback - code should be wayyy simpler now :D

yoshuawuyts:
@juliangruber thanks for the quick review!

PR #110, opened 2016-12-09
Refactor component styles

Kriesse:
Removed scss where possible, styles now live in the element as either tachyons classes or custom classes via sheetify.


Kriesse:
@yoshuawuyts SCSS is almost ready to go – need help with two things:
* the progress bar should be its own component but wasn't sure how to get it out of table.js, can you help?
* not sure where the styles in base.scss should live.

yoshuawuyts:
Quick comment: in the readme the `"build assets"` section should probs also be updated

elements/dat-import.js
```diff
@@ -0,0 +1,85 @@
...
+  }
+
+  .icon-link {
+    padding-top: .32rem;
+    padding-left: .5rem;
+    pointer-events: none;
+    width: var(--icon-height);
+    height: var(--icon-height);
+  }
+  input {
```
yoshuawuyts: beware that this will style __all__ `<input>` fields. If you wanna make it private, prefix it with `:host > input` or something

elements/modal.js
```diff
@@ -164,32 +145,32 @@ function createWidget () {
...
-              <use xlink:href="#daticon-clipboard" />
-            </svg>
+          <input name="dat-link" type="text" value=${link} class="relative dib pa0 dat-input-input">
+          ${icon({
+            id: 'link'
+          })}
+          <button class="absolute pointer dat-input-button">
+            ${icon({
+              id: 'clipboard'
+            })}
```
yoshuawuyts: I think if all these were on a single line it'd be slightly more readable (:

yoshuawuyts:
Yeah feel this looks mostly alright; I'm slightly concerned about conflicts in class names tho; using the `:host` selector to prefix the classes might help. [Here are the docs in sheetify](https://github.com/stackcss/sheetify#styling-host-elements); I'm not sure if it's clear enough that this might still lead to conflicts?

yoshuawuyts:
@Kriesse re your questions:

I think if you replace the SASS vars with CSS variables, and like put them in a `./base.css` file or something we could import it from our `app.js` file doing:
```js
css('./base.css')
```

I think it's fine if you inline the progress bar CSS into `table.js` for now and then I can help with extracting it from there. Does that sound alright?

Kriesse:
@yoshuawuyts sounds good, will address all of this!

Re: the namespacing using `:host`:
That’s what I did orignally, but I found that even without explicitly using `:host`, the selector will be prefixed. 
Can you reproduce? Not sure this is a sheetify bug, or intentional?

image

yoshuawuyts:
@Kriesse ooooops, we're running an old version of sheetify 😱

yoshuawuyts:
https://github.com/datproject/dat-desktop/pull/112 now exists - try
checking it out; stuff is def breaking

Kriesse:
@yoshuawuyts prefixed all the things.

Attempted to add global styles using `css('./public/css/base.css', { global: true })`, but it's not working yet.

However, how about we merge this and https://github.com/datproject/dat-desktop/pull/112/ and continue from there?

yoshuawuyts:
@Kriesse yeah sounds good to me!

PR #111, opened 2016-12-10
add unstyled welcome screen

yoshuawuyts:
- renamed `models/app` to `models/repos`
- created a `main-view` model used to handle the state from `views/main`
- added `choo-persist` v2 that auto-persists all state except `models/repos` to `IndexedDB`. `models/repos` cannot be persisted because the values inside it cannot be serialized
- created a welcome screen that's shown the first time you boot up the application - didn't do any styling so @Kriesse can do her thing

---
## See Also
- https://github.com/datproject/dat-desktop/issues/68

---
## Screenshot

screen shot 2016-12-10 at 01 52 01

### DIFF
--- a/app.js
+++ b/app.js
@@ -1,15 +1,24 @@
+const persist = require('choo-persist')
 const mount = require('choo/mount')
 const log = require('choo-log')
+const xtend = require('xtend')
 const choo = require('choo')
 
-const mainView = require('./pages/main')
+const opts = {
+  filter: (state) => {
+    state = xtend(state)
+    delete state.app
+    return state
+  }
+}
 
-const app = choo()
-app.use(log())
+persist(opts, (p) => {
+  const app = choo()
+  app.use(log())
+  app.use(p)
 
-// import & init models
-app.model(require('./models/app')())
+  app.model(require('./models/app')())
 
-// start
-app.router(['/', mainView])
-mount('body', app.start())
+  app.router(['/', require('./pages/main')])
+  mount('body', app.start())
+})
--- a/app.js
+++ b/app.js
@@ -7,7 +7,7 @@ const choo = require('choo')
 const opts = {
   filter: (state) => {
     state = xtend(state)
-    delete state.app
+    delete state.repos
     return state
   }
 }
@@ -17,7 +17,8 @@ persist(opts, (p) => {
   app.use(log())
   app.use(p)
 
-  app.model(require('./models/app')())
+  app.model(require('./models/main-view')())
+  app.model(require('./models/repos')())
 
   app.router(['/', require('./pages/main')])
   mount('body', app.start())
--- a/models/repos.js
+++ b/models/repos.js
@@ -9,13 +9,13 @@ const Manager = require('../lib/dat-manager')
 module.exports = createModel
 
 function createModel (cb) {
-  const model = Model('app')
+  const model = Model('repos')
 
   const manager = new Manager()
 
   model.subscription('manager', (send, done) => {
     manager.on('update', () => {
-      send('app:updateArchives', manager.get(), done)
+      send('repos:updateArchives', manager.get(), done)
     })
   })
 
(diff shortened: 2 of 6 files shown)


pages/main.js
```diff
@@ -113,6 +124,16 @@ function mainView (state, prev, send) {
   `
 }
 
+function WelcomeScreen (methods) {
+  const onExit = methods.onexit
```
yoshuawuyts: I chose to not use camelcase here to keep the method names similar to `lib/modal` which in turn is similar to `onload` which in turn is similar to the built-in DOM events

yoshuawuyts:
Realized we shouldn't be showing the header. Updated, it's now a proper clean slate :tada: :

screen shot 2016-12-10 at 02 05 25

app.js
```diff
@@ -1,15 +1,25 @@
+const persist = require('choo-persist')
 const mount = require('choo/mount')
 const log = require('choo-log')
+const xtend = require('xtend')
 const choo = require('choo')
 
-const mainView = require('./pages/main')
+const opts = {
+  filter: (state) => {
+    state = xtend(state)
```
juliangruber: simpler and es6: `state = Object.assign({}, state)`

models/repos.js
```diff
@@ -9,13 +9,13 @@ const Manager = require('../lib/dat-manager')
...
 function createModel (cb) {
-  const model = Model('app')
+  const model = Model('repos')
 
   const manager = new Manager()
 
   model.subscription('manager', (send, done) => {
     manager.on('update', () => {
-      send('app:updateArchives', manager.get(), done)
+      send('repos:updateArchives', manager.get(), done)
```
juliangruber: this event uses two different words for the same concept, can we rename this to say `repos:update`?

pages/main.js
```diff
@@ -75,14 +75,24 @@ module.exports = mainView
 // render the main view
 // (obj, obj, fn) -> html
 function mainView (state, prev, send) {
+  const showWelcomeScreen = state.mainView.showWelcomeScreen
   const modalLink = state.location.search.modal
-  const dats = state.app.archives
+  const dats = state.repos.archives
```
juliangruber: same about the wording here

pages/main.js
juliangruber: +1

app.js
yoshuawuyts: I think while your suggestion might be easier, it's not simpler - I don't like `Object.assign()` for the reason that it doesn't work everywhere and polyfills are large; I'd rather leave `extend` in there since we're already using it in other places too. I'd suggest we moderate ourselves with the usage of ES6 given most of the team barely uses any of it also

yoshuawuyts:
Updated with all feedback! cc/ @juliangruber :sparkles:

app.js
juliangruber: 1) It works in electron -> we don't need a polyfill
2) It will eventually work everywhere
3) It helps keep the dependency tree a bit simpler without extra lines of code
4) It will become the de facto standard for cloning - and I'd rather we're on the forefront of adoption
5) Better to teach and learn about it now than later
6) It's part of the programming language we're using

juliangruber: 7. We should get rid of `xtend` in the other places and then remove the dependency completely

karissa: whats wrong with extend? i'm confused

juliangruber: @karissa `Object.assign()` is JavaScript's own solution to this problem. It's part of the language's core and should therefore be preferred, especially in environments like Electron where you know exactly that you'll get an up to date version of V8.

yoshuawuyts: I'd rather we don't get hung up on syntax, including making people learn things they don't otherwise know. Like every PR I make there's comments about syntax or new specs and I feel it slows us down - I don't wanna learn the quirks of classes or details of `Object.extend()` when we could be improving our design and fixing bugs instead.

I'm happy to have a convo about syntax and language features, but I don't feel it's enough to stall a PR over. Let's have that convo another time, perhaps in person when we hang out; think that'd work better than discuss it over text (:

juliangruber: Yeah, I agree about not getting hung up on this now. But I don't get how people aren't interested in learning more about the language they use.

PR #127, opened 2017-01-03
style welcome page #68

Kriesse:


pages/main.js
```diff
@@ -63,6 +64,19 @@ const prefix = css`
...
+  :host {
+    height: 100vh;
+    background-color: var(--color-neutral);
+    color: var(--color-white);
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+    justify-content: center;
+    text-align: center;
+  }
```
yoshuawuyts: This can be replaced with the tachyons classes of:
```txt
100-vh flex flex-column align-center justify-center tc
```
The only value not included is `var(--color-neutral)` but I believe we had a class for this already. This saves us 10 lines of code in the codebase

PR #129, opened 2017-01-08
use open to open files

yoshuawuyts:
Uses `opener` for cross-platform file opening compat. Had to fork it because of an issue with the Acorn parser, which is solved by https://github.com/domenic/opener/pull/22 - I've pointed it to our fork in the mean time. Thanks! :sparkles:

### DIFF
--- a/models/repos.js
+++ b/models/repos.js
@@ -1,8 +1,8 @@
 const dialog = require('electron').remote.dialog
 const ipc = require('electron').ipcRenderer
-const exec = require('child_process').exec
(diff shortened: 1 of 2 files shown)


yoshuawuyts:
Apparently the `opener` author isn't concerned with parsers other than v8, so switching to `open` instead. Yay, tested it and PR should be all good now! :sparkles:

juliangruber:
very nice! :)

karissa:
nice job

Issue #69, opened 2016-10-21
[infrastructure] telemetry


yoshuawuyts:
In order to know how Dat desktop is performing, it would be useful to create a telemetry dashboard similar to how webtorrent-desktop does it. Stats would include user retention rate, common bugs, and more.
### steps
- [ ] extract telemetry code from [webtorrent-desktop](https://github.com/feross/webtorrent-desktop) (https://github.com/feross/webtorrent-desktop/blob/master/src/renderer/lib/telemetry.js)
- [ ] extract telemetry UI code from [webtorrent-www](https://github.com/feross/webtorrent-www)
  - [ ] https://github.com/feross/webtorrent-www/blob/master/bin/summarize-telemetry.js
  - [ ] https://github.com/feross/webtorrent-www/blob/master/server/desktop-api.js
  - [ ] https://github.com/feross/webtorrent-www/blob/master/server/views/telemetry-dashboard.pug
- [ ] deploy telemetry UI code
- [ ] integrate telemetry code with `dat-desktop`

PR #130, opened 2017-01-09
Warn uncaught error

### LINKED ISSUE #123: debug dialog with uncaught errors
tell people something sane so they don't get too frustrated on errors

### LINKED ISSUE #124: error dialog
tell people something sane so they don't get too frustrated on errors

yoshuawuyts:
There's a lot going on in this patch, but the changes are relatively straight forward and each one relies on the previous one so I decided to split them by commit rather than creating a new PR for every commit. In this patch we:

- split out the main view into separate views for each of the querystring states; so we have a `main-share`, `main-delete`, etc.
- fixed an error in the delete modal. It wasn't actually deleting stuff it seems; not sure when the regression happened but it's fixed now
- create a modal when an error occurs on the renderer thread. We don't do any logic on the main thread so this should be sufficient to guard us from most runtime errors

Hope this explains what's going on. Closes #123 . Thanks!

__Note for reviewers:__ I tried hard to keep the code consistent with the project style - hope it's close enough (:

----
### runtime error screenshot
screen shot 2017-01-09 at 15 37 29


yoshuawuyts:
oops, mislabeled folks for review - anyone's welcome to review though :D - using explicit tags just so certain folks get notified (:

elements/basic-modal.js
```diff
@@ -0,0 +1,72 @@
...
+      }
+    }
+  })
+
+  function render (onOk, onExit) {
+    return html`
+      <section class="relative flex flex-column justify-center ${prefix}">
+        <h3 class="f4">Unexpected issue</h3>
+        <p class="mt3 mb4 f7 color-neutral-70">
+          There was an unexpected issue. We will need to close the app. Sorry
```
karissa: Can we tell them to join IRC? Is there any way to get the stack trace they can copy?  Perhaps for a separate PR !

yoshuawuyts:
@karissa yeah ideally we'd have a thing where the error is uploaded to one
of our servers so we can inspect it and debug it. That'd be part of
https://github.com/datproject/dat-desktop/issues/69 then I think

app.js
```diff
@@ -1,11 +1,16 @@
 const persist = require('choo-persist')
 const mount = require('choo/mount')
 const log = require('choo-log')
+const css = require('sheetify')
```
Kriesse: Just curious so I know for the future: What's the convention you follow for the order and naming of variables here?

karissa:
Sending an error to servers sounds like a different (more complicated) thing -- I was simply thinking of an input box with the text there so that the user could copy/paste it to us! totally unnecessary right now but might be a nice feature

Kriesse:
@karissa @yoshuawuyts Is this ok to merge, and we can track crash reporting in #69 (or a new separate issue)?

app.js
yoshuawuyts: @Kriesse I like the words `css` and `html` for `sheetify` and `bel` respectively because github detects these and applies syntax highlighting. The rest is usually camelcased stuff (:

Kriesse: @yoshuawuyts got it, thanks :)

PR #154, opened 2017-01-18
Fix browserify bundling

juliangruber:
`__dirname` is now being resolved correctly, so it will work on any machine and in the packaged app as well. Also as a benefit not the `"bundle"` and the `"watch"` scripts always use the same browserify arguments.


juliangruber:
yosh confirmed it works!

PR #158, opened 2017-01-20
burnthemall: remove redundant rebuild step

juliangruber:
the postinstall script makes sure rebuild is run already


karissa:
fixed on master

PR #161, opened 2017-01-22
Fix modals

### LINKED ISSUE #74: npm install should install electron native modules correctly first time
based on reading https://docs.npmjs.com/all#config i was hoping adding this to dat-desktop package.json would make npm install "just work" but it had no effect :( https://usercontent.irccloud-cdn.com/file/AoNdfyK3/Screen%20Shot%202016-11-02%20at%201.57.32%20PM.png

to fix it, @juliangruber came up with the idea of having `prebuild` respect the `config` property of package.json, which would solve the problem of needing a `npm run rebuild` script

### LINKED ISSUE #140: screen flickers when replicating dat while share dialogue is open

yoshuawuyts:
- added a debounce function to the `repos:update` call. This was necessary to kick off debugging our modal problems
- fixed modal flickering after copying a link. Closes #140 
- modals can now be closed by `esc` or clicking outside the frame. Closes #74 
- migrated all modal code to [base-elements/modal](https://github.com/yoshuawuyts/base-elements#modalopts), which makes it a lot nicer to maintain (and allowed squashing the above bugs)
- removed `lib/modal-element`

Thanks!


juliangruber:
Hope you don't mind my slight style modernizations..

LGTM!

I'm curious, what specifically was causing the modal flickering?

yoshuawuyts:
@juliangruber hah, all cool (:

Glad you like it. The problem was that we were nesting two instances of `un-load` and with the re-render step it would load and then unload the content somehow. E.g. the API was messy and not thought out well enough. Doing a bunch of work on `nanocomponent` helped in building like a better API - something I think we can probably leverage for other components tooooo :sparkles:

Issue #152, opened 2017-01-17
Improve renderer performance


yoshuawuyts:
## todos
- [ ] cache elements (#250)
- [ ] switch off `morphdom` to `nanomorph` (https://github.com/yoshuawuyts/choo/issues/152)

---
## Issue
Looks like doing renders in quick succession means the app starts binging on resources. We've got some solid offenders in there:
- __12%__ the `copier` function from `yo-yo` to copy events. Can be fixed by switching to `nanomorph`
- __25%__ `log`, `groupCollapsed` and `groupEnd` are all part of choo-log. Given at a 60fps refresh rate they're called each time they tend to eat a lot of cycles. We should um, definitely disable this in production but also think about how to improve perf. Perhaps a debounce on stuff that triggers the 60fps? Suggestions welcome!
- __18%__ `morph` - apparently morphing stuff takes a lot of time. Using `nanomorph` + caching elements through `nanocomponent` should probably help with this, as it means we can do partial recomputes of the tree rather than the whole thing, each time

This accounts for over half of performance stuff; if we squash these we're 2x as fast as performant. We probably also always want to run `yo-yoify` as there's a scattershot of usage in the upper regions related to not using it, it seems.

<img width="658" alt="screen shot 2017-01-17 at 16 50 34" src="https://cloud.githubusercontent.com/assets/2467194/22028137/809178b6-dcd6-11e6-8d56-919692cd6210.png">

juliangruber:
> We should um, definitely disable this in production but also think about how to improve perf. Perhaps a debounce on stuff that triggers the 60fps? Suggestions welcome!

Debouncing page updates sounds good to me!

Shall we create individual issues for each mentioned optimization?

yoshuawuyts:
juliangruber, yeah if you want to - I'm not necessarily fussed about
splitting these into multiple issues because each of these should be fairly
easy to resolve? But that's me, no objections if you want to split 'em tho
(:

yoshuawuyts:
Implementing debounce in #161 - has limited effects, but useful regardless. I think having a multidrive / dat-next refactor will help improve this even more.

yoshuawuyts:
Marked as priority high b/c if the app makes folks CPUs spin there's a fair chance they'll choose not to use it

yoshuawuyts:
Updating to nanomorph is probably going to be part of `choo@5` https://github.com/yoshuawuyts/choo/issues/152 - blocked on that

[...]

PR #165, opened 2017-01-23
only run choo-log in development

yoshuawuyts:
Should help boost performance a lil; ref https://github.com/datproject/dat-desktop/issues/152
- start using `envify` to hardcode strings for production
- only run `choo-log` if `NODE_ENV` is development

### DIFF
--- a/app.js
+++ b/app.js
@@ -22,8 +22,11 @@ const opts = {
 
 persist(opts, (p) => {
   const app = choo()
-  app.use(log())
   app.use(p)
+
+  if (process.env.NODE_ENV === 'development') {
+    app.use(log())
+  }
 
   app.model(require('./models/main-view')())
   app.model(require('./models/window')())
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "dependency-check": "^2.6.0",
     "electron": "1.4.14",
     "electron-builder": "^8.6.0",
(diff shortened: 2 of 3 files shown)


app.js
```diff
@@ -22,9 +21,13 @@ const opts = {
 
 persist(opts, (p) => {
   const app = choo()
-  app.use(log())
   app.use(p)
 
+  if (process.env.NODE_ENV === 'development') {
+    const log = require('choo-log')
```
juliangruber: would it be bad to require this module anyway? would be cleaner to have all the requires at the top of the file

package.json
```diff
@@ -6,6 +6,11 @@
   "repository": "datproject/dat-desktop",
   "description": "Dat Desktop App",
   "author": "Dat Team",
+  "browserify": {
+    "transform": [
+      "envify"
+    ]
+  },
```
juliangruber: should be moved to `scripts/browserify.js` for cohesiveness

yoshuawuyts: Oh, I wasn't aware we had that now. Sure (:

app.js
yoshuawuyts: Oh yeah, given it's a desktop app this should be ok

yoshuawuyts:
@juliangruber see #152

juliangruber:
awesomeeeeee!

PR #169, opened 2017-01-24
fix sodium-prebuilt using updated bindingify

juliangruber:
@Kriesse can you give this a try? This should fix the sodium errors without special casing any dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,7 +58,6 @@
     "dist": "build"
   },
   "devDependencies": {
-    "bindingify": "yoshuawuyts/bindingify#fix/sodium",
     "browserify": "^13.3.0",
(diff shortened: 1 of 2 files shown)


juliangruber:
@yoshuawuyts i was able to simplify this by _a lot_ by simply only bundling our own files and sheetify, the rest is ignored 😎

juliangruber:
this brings build times down from ~10s to ~2s on my machine!

juliangruber:
> Do module specific transforms still work tho? Not a blocker, mostly just curious

what do you mean by that?

juliangruber:
this bypasses browserify _completely_ for everything except our code and sheetify and its deps

PR #170, opened 2017-01-26
scripts/build: remove --build-from-source

juliangruber:
Originally introduced to fix native dependency issues, we found
out that browserify configuration was to blame instead. So this
enables prebuild installs again!


karissa:
Woot

PR #172, opened 2017-01-26
models/error: properly rethrow error, preserving stack

juliangruber:
this way we get the clickable code locations in the dev tools

### DIFF
--- a/models/error.js
+++ b/models/error.js
@@ -18,6 +18,8 @@ function model () {
 
 function onUncaughtException (send, done) {
   process.on('uncaughtException', function (err) {
+    if (err._thrown) return
+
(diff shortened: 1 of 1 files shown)


models/error.js
```diff
@@ -18,6 +18,10 @@ function model () {
 
 function onUncaughtException (send, done) {
   process.on('uncaughtException', function (err) {
+    if (err._thrown) return
+    err._thrown = true
+    throw err
```
yoshuawuyts: oh actually this makes standard fail - should go down further (:

yoshuawuyts:
pushed a fix, think this should help (:

PR #173, opened 2017-01-26
Upgrade to dat-node@1. Fixes #168

### LINKED ISSUE #168: upgrade to dat-node@latest
for #167

juliangruber:
Major changes:

- moved from emitting events on the dat instance, which isn't an eventemitter any more, to an emitter at `dat.events`
- no need to persist `dat.owner` any more
- make `dat.stats` a getter to the underlying hyperstats instance
- no need to manually close `dat.db()` any more


lib/dat-manager.js
```diff
@@ -19,18 +18,22 @@ module.exports = class Manager extends EventEmitter {
...
   }
   create (dir, cb) {
-    const dat = Dat({ dir })
-    dat.owner = true
-    dat.dir = dir
-    dat.share(err => {
+    Dat(dir, (err, dat) => {
       if (err) return cb(err)
+
+      dat.dir = dir
```
joehand: Oops sorry changed API there. I can expose this if that'd be helpful. `dat.path` is the full resolved path but `dat.dir` could be the relative `dir` passed in.

lib/dat-manager.js
```diff
@@ -44,102 +47,102 @@ module.exports = class Manager extends EventEmitter {
...
-          }
-          dat.title = json.title
-          dat.author = json.author
-          dat.emit('update')
-        })
+    dat.events = new EventEmitter()
+    dat.listStream = dat.archive.list({ live: true })
+    dat.listStream.on('data', entry => {
+      if (entry.name !== 'dat.json') return
+      collect(dat.archive.createFileReadStream('dat.json'), (err, raw) => {
```
joehand: Ooh nice. So if there is an update to `dat.json` (i.e. more than one `dat.json` entry in metadata), the UI will just keep updating until the last metadata entry?

lib/dat-manager.js
juliangruber: oh, i wasn't aware of `dat.path`. I was adding `dat.dir` manually before I think, I can just use `dat.path` then

joehand: 👍  the only difference is `dat.path` will be fully resolved, not sure if `dir` was there.

juliangruber: addressed

lib/dat-manager.js
juliangruber: yup currently it's doing redundant work. Could we make it so the "live" stream only starts at the current point in time and goes forward?

juliangruber:
thanks for the review so far, @joehand!

lib/dat-manager.js
joehand: No just wanted to confirm the UI would update. If it starts at current point you may not get the `dat.json`, right? (if its the first entry) Or maybe it'd be persisted in the db already?

Getting one file will be much easier with the append-tree update from mafintosh - so this probably is good for now.

joehand:
Looks good! Hopefully it was relatively seamless to use the new dat-node 😬

lib/dat-manager.js
juliangruber: oh, yeah the ui will update, that's the `this.emit('update')` call underneath

juliangruber:
@joehand after the versioning confusing yes it was :)

I would just also keep master the current stable version, otherwise it's confusing to look at the repo's code and see a different version from what you're using.

yoshuawuyts:
We found multiple regressions, the first was with importing a dat and then reopening the window it would crash because `.dir` was renamed to `.path`. We've fixed this with the changes we've pushed

The second issue is that we cannot delete dats that we create and the window freezes without throwing an error. We created a dat that was ~50mb. We weren't able to fix this but tracked it down to the `remove` method in `data-manager`. `dat.close()` never calls a callback (cc/ @joehand)

screen shot 2017-01-27 at 12 45 25

juliangruber:
> We found multiple regressions, the first was with importing a dat and then reopening the window it would crash because .dir was renamed to .path. We've fixed this with the changes we've pushed

Ah, thanks for catching this bug and fixing it!

> dat.close() never calls a callback (cc/ @joehand)

I've seen this too, in https://github.com/datproject/dat-node/issues/84. @joehand could you run the app and see if you can reproduce? Otherwise we could also try to create a node only gist of this happening.

elements/table.js
```diff
@@ -195,15 +195,12 @@ function tableElement (dats, send) {
...
   return dats.map(dat => {
-    const stats = dat.hyperStats && dat.hyperStats.get()
-    const progress = stats
-      ? stats.blocksTotal
-        ? Math.round(stats.blocksProgress / stats.blocksTotal * 100)
-        : 100
-      : 0
+    const progress = dat.stats.blocksTotal
+      ? Math.round(dat.stats.blocksProgress / dat.stats.blocksTotal * 100)
+      : 100
```
juliangruber: quick fix!

PR #174, opened 2017-01-26
Include meta in total progress. Closes #166

### LINKED ISSUE #166: UI for syncing metadata
On _very large dats™_ it can take a while to sync the metadata. This happens before any chunks get downloaded, so the progress stays at 0% for a while. This is why in the CLI we show two progress bars. I was wondering if we could perhaps come up with a good way to signal this in the desktop too.

cc/ @Kriesse got any ideas for this perhaps?

---
### cli screenshot
screen shot 2017-01-23 at 02 21 37

juliangruber:
This includes meta download progress in total progress. Code inspired by dat-next.

### DIFF
--- a/lib/dat-manager.js
+++ b/lib/dat-manager.js
@@ -111,6 +111,16 @@ module.exports = class Manager extends EventEmitter {
     }, cb)
   }
   _activate (dat) {
+    Object.defineProperty(dat, 'progress', {
+      get: () => {
+        const stats = dat.hyperStats && dat.hyperStats.get()
+        return stats
+          ? stats.blocksTotal
+            ? Math.round(stats.blocksProgress / stats.blocksTotal * 100)
+            : 100
+          : 0
+      }
+    })
     dat.open(err => {
       if (err) throw err
       dat.listStream = dat.archive.list({ live: true })
--- a/lib/dat-manager.js
+++ b/lib/dat-manager.js
@@ -114,11 +114,20 @@ module.exports = class Manager extends EventEmitter {
     Object.defineProperty(dat, 'progress', {
       get: () => {
         const stats = dat.hyperStats && dat.hyperStats.get()
-        return stats
-          ? stats.blocksTotal
-            ? Math.round(stats.blocksProgress / stats.blocksTotal * 100)
-            : 100
-          : 0
+
+        let progress = 0
+        let total = 0
+
+        if (dat.archive) {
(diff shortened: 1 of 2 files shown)


lib/dat-manager.js
```diff
@@ -111,6 +111,25 @@ module.exports = class Manager extends EventEmitter {
...
+
+        let progress = 0
+        let total = 0
+
+        if (dat.archive) {
+          progress += dat.archive.metadata.blocks - dat.archive.metadata.blocksRemaining()
+          total += dat.archive.metadata.blocks
+        }
+        if (stats) {
+          progress += stats.blocksProgress
```
joehand: FYI been seeing a but with hyperdrive stats `blocksProgress` being greater than `blocksTotal`. I have a [workaround in dat-next](https://github.com/joehand/dat-next/blob/master/lib/download.js#L109-L112). That I think solves it. But I'm pretty sure its the same calculation so I have no idea what is going on. 

Sorry haven't had time to chase that down...

joehand: pfrazee also had some content progress issues with similar calculation but may have been solved with this pr: https://github.com/mafintosh/hyperdrive/pull/123. 

The issue was that some clients don't want the historic blocks so they never "finish".

juliangruber:
@joehand what about this as a simple fix for now?

joehand:
> @joehand what about this as a simple fix for now?

Works for now probably. I'm gonna chase this bug next week first thing!

This still may show as 100% done even if it is not (sometimes I had to get to 150% for it to be 'done'). So as long as you aren't stopping anything with 100% - that was my issue with the CLI was I exited at 100% even though that wasn't finished.

juliangruber:
I think what you're saying is also in the case where `*Total` isn't fully known yet it will mistakenly show `100%`, right?

Yeah, we're not triggering anything once 100% is reached, it's purely for UI.

yoshuawuyts:
This isn't running for me:

screen shot 2017-01-30 at 15 53 26

juliangruber:
looks like you need to reinstall dat-node and bundle

juliangruber:
ah, no! @yoshuawuyts this is because your PR fixing dat-node changed the DB schema, you need to wipe your db

yoshuawuyts:
This has become part of #198, files referenced no longer exist - closing! :sparkles:

PR #177, opened 2017-01-26
[wip] add integration tests

yoshuawuyts:
for https://github.com/datproject/dat-desktop/issues/155 - like with WebTorrent we could make this a thing we run before publishing; not necessarily needed to automate on travis and stuff yet given that `xvfb` can be a pain to setup


juliangruber:
Nice!

travis shouldn't be too hard to set up, just needs this: https://github.com/juliangruber/browser-run#headless-testing

tests/index.js
```diff
@@ -0,0 +1,114 @@
...
+  //     .then(() => waitForLoad(app))
+  //     .then(() => app.browserWindow.isVisible())
+  //     .then((isVisible) => t.equal(isVisible, true))
+  //     .then(() => app.client.click('button'))
+  //     .then(() => app.browserWindow.getTitle())
+  //     .then((title) => t.equal(title, 'Dat Desktop'))
+  //     .then(() => endTest(app, t), (err) => endTest(app, t, err || 'error'))
+  // })
+
+  t.test('make sure you can minimize, full screen, resize, and move the window')
```
karissa: probably don't need to test window resizing. more important to test key features that will prevent folks from using the app/ the app breaking entirely

karissa:
OK this is now mergeable and should be ready (pass locally)

models/repos.js
```diff
@@ -11,6 +11,11 @@ const assert = require('assert')
 const mkdirp = require('mkdirp')
 const open = require('open')
 const path = require('path')
+const mock = require('../tests/mocks')
+
+if (process.env.RUNNING_IN_SPECTRON) {
+  mock(dialog)
+}
```
juliangruber: Requiring a file from the tests dir in lib code feels weird, we might for example not want to ship the tests dir with the production bundle. Also, someone could introduce a dependency in the `tests/mocks` file but only add it to `devDependencies`.

Do you maybe see another way to resolve this? Maybe even with a lib that patches `require()`?

karissa: Makes sense, I'm not sure how to make this better though. If you have an idea that'd be useful but I think that i'd rather require something from tests than patch require.

juliangruber: oh, what about we move that mock function to `lib/`, as a simple solution?

karissa:
Seems like it's failing only on linux.. why is that?

juliangruber:
I haven't run the app on linux before, so far we only tested whether it builds there. We could just remove linux from travis until we consider it a deployment target.

juliangruber:
(it might also be a bug in the integration libs though of course)

karissa:
Yeah, let's remove it from the build until it's a deployment target

PR #189, opened 2017-01-31
Add in-process bundling. Closes #187

### LINKED ISSUE #187: Run the `watch` task in process
This way it's always keeping bundles up to date in dev, and no need to keep a second terminal open

juliangruber:


index.js
```diff
@@ -43,7 +43,7 @@ menu[menu.length - 1].submenu.push({
   }
 })
 
-app.on('ready', () => {
+const onReady = () => {
```
yoshuawuyts: Minor nit: could we make this a normal function if possible? I'm cool if you like arrow functions but I always find these `var foo = function() {}` style functions super hard to read

yoshuawuyts:
Code looks good to me. I think once the conflict is resolved we can def merge it :sparkles:.

I was thinking tho: would it perhaps make sense to start moving onto `bankai`? - right now this is turning out to be a reasonable amount of project specific code and moving it out of the project might help keep our codebase clean. It would require a few tweaks to `bankai` tho, (e.g. https://github.com/yoshuawuyts/bankai/issues/123) but reckon it might be worth it. Thoughts?

index.js
juliangruber: i'm cool with that too, will change

juliangruber:
I agree that it would make sense to move to bankai! After public launch maybe?

yoshuawuyts:
woot, yeah sounds good :sparkles:

PR #190, opened 2017-01-31
fix bundle if dat-desktop parent is node_modules folder

joehand:
My bundle was not working because the path of the repo included `node_modules`. Make sure we only ignore node_module files relative to the repo directory.


juliangruber:
nice, thanks!

PR #192, opened 2017-02-01
Do not capitalize author

joehand:
Doing text transform capitalize can lead to weird author things like: 

screen shot 2017-02-01 at 11 57 30

Also some people like their name lower case. We should use whatever the author specified.


karissa:
+1

yoshuawuyts:
Fair enough (:

PR #193, opened 2017-02-01
remove choo-model

yoshuawuyts:
realized `choo-model` is not ideal hah; just moving some code around (:

### DIFF
--- a/models/main-view.js
+++ b/models/main-view.js
@@ -1,14 +1,17 @@
-const Model = require('choo-model')
-
 module.exports = createModel
 
 function createModel (cb) {
-  const model = Model('mainView')
+  return {
+    namespace: 'mainView',
+    state: {
+      showWelcomeScreen: true
+    },
+    reducers: {
+      showWelcomeScreen: showWelcomeScreen
+    }
+  }
+}
 
-  model.state({ showWelcomeScreen: true })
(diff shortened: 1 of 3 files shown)


juliangruber:
to me this looks cleaner with choo-model, what's the issue with it?

karissa:
love all the work you're putting into making the codebase stable and awesome! I'm down to merge this, but in general I always wonder if we're refactoring things prematurely -- i want to make sure we can roll out the desktop app with all the goodies before it's all old news!! :)

karissa:
(and we're ALL guilty of this!!)

yoshuawuyts:
@juliangruber got seriously frustrated the other day when refactoring all the `modal` stuff and trying to figure out which events were available. Thought `choo-model` was going to help with readability but instead it's been doing the exact opposite for me. By not using `choo-model` we got a manifest of available methods available on the model which is useful when you're trying to use the API. Also makes it easier to spot when our API is complicated or weirdly named so we can later go in and change it.

@karissa took me 10 mins to whip this out, I generally agree with you but figured this might be a quick easy win (:

karissa:
Cool beans!

juliangruber:
@yoshuawuyts ok +1 then, if we want we can think about a better abstraction later then

PR #195, opened 2017-02-03
add soon required insert-css dependency

juliangruber:
npm3 doesn't automatically install peer dependencies any more. we're currently getting a warning about this.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,6 +28,7 @@
     "explain-error": "^1.0.3",
     "global": "^4.3.1",
     "hyperdrive-stats": "^2.2.5",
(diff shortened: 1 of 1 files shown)


yoshuawuyts:
eh, I don't know why this would pop up. We removed the peer dep requirement in `sheetify@6` by shipping [insert.js](https://github.com/stackcss/sheetify/blob/master/insert.js). What error are you seeing?

juliangruber:
base-elements is still using `sheetify@5`, that seems to be it

yoshuawuyts:
published `base-elements@2.2.0` :tada:

PR #198, opened 2017-02-04
Multidat

yoshuawuyts:
This touches some of the same code as https://github.com/datproject/dat-desktop/pull/173. I think once this patch is merged we can move to some of the ergonomics #173 provides (e.g. using getters and the like for state access). This patch should provide us with some robust underpinnings to work on top :sparkles:

### changes
- removes all dependencies on `level` which preps us for the upcoming `SLEEP` release
- moved state into `~/.dat-desktop-db.json` using `toiletdb`
- renamed some methods on `model/repos` to be more in line with the dat cli
- moved our custom `dat` management code into [multidat](https://github.com/yoshuawuyts/multidat) library. It's fully tested to make sure it works as expected and relies on [multidrive](https://github.com/yoshuawuyts/multidrive) and the latest `dat-node`
- made all unhandled errors call back instead. Might lead to crashes where errors were previously swallowed
- added a test to catch unused deps
- removed a whole bunch of unused deps
- removed more code than was added which is always :cool:

### todo
- [x] use the `multidat.readManifest()` method to determine author and title
- [x] fix key restart regression in https://github.com/datproject/dat-node/commit/508de80063bf87b90e4142bf5803b7d5d5d23c21

### notes
- placing this under the public launch milestone because moving off level should be done sooner rather than later
- I thought #173 was merged already before submitting this PR; apologies that there's some overlap there!
- during the writing of this patch we found a few subtle bugs in `dat-node`; they're solved now but we'll probably run into more
- there's less DB lock errors it seems, but we're hitting different state corruptions now; I think we'll need to focus on debugging for the next week or so to have it stable enough for a release


lib/root-dir.js
```diff
@@ -1,8 +0,0 @@
-const {app, process: remoteProcess} = require('electron').remote
-const minimist = require('minimist')
-const fs = require('fs')
-
-const argv = minimist(remoteProcess.argv.slice(2))
-const rootDir = argv.data || `${app.getPath('downloads')}/dat`
```
juliangruber: can we add this back? this is required so you can run two instances of dat-desktop on the same machine, to verify they're able to replicate with each other properly

models/repos.js
```diff
@@ -1,109 +1,205 @@
...
-  function handleManagerUpdate (send, done) {
-    manager.on('update', debounce(() => {
-      send('repos:update', manager.get(), done)
-    }), 1500, true)
+  // boot multidat, create the ~/Downloads/dat directory
+  function startMultidat (send, done) {
+    var dblocation = path.join(process.env.HOME)
+    var dbfile = path.join(dblocation, '.dat-desktop-db.json')
+
+    var arr = [
```
juliangruber: this variable name is a bit generic, can we change this to `waterfall([fun...` or `var tasks`?

models/repos.js
```diff
@@ -1,109 +1,205 @@
...
     }
   }
 
-  function handleManagerUpdate (send, done) {
-    manager.on('update', debounce(() => {
-      send('repos:update', manager.get(), done)
-    }), 1500, true)
+  // boot multidat, create the ~/Downloads/dat directory
+  function startMultidat (send, done) {
+    var dblocation = path.join(process.env.HOME)
```
juliangruber: should be camelCased, to be consistent

models/repos.js
```diff
@@ -1,109 +1,205 @@
...
+    var dbfile = path.join(dblocation, '.dat-desktop-db.json')
+
+    var arr = [
+      function (next) {
+        mkdirp(dblocation, next)
+      },
+      function (val, next) {
```
juliangruber: we're not using `val`, can you rename it to `_`?
also the `val` below would be more clear when renamed to `multidat`, as that's the identifier used below.

models/repos.js
```diff
@@ -1,109 +1,205 @@
...
+      if (!files || !files.length) return done()
+      pathname = files[0]
+    }
+    manager.create(pathname, done)
+  }
+
+  function removeDat (state, data, send, done) {
+    assert.ok(data.key, 'repos-model.deleteDat: data.key should exist')
+
+    if (data.confirmed === true) {
```
juliangruber: because the case where `data.confirmed === false` always happens first, can we list it first here too?

models/repos.js
```diff
@@ -1,109 +1,205 @@
...
+
+  // copy a dat share link to clipboard and open a modal
+  function shareDat (state, data, send, done) {
+    assert.ok(data.key, 'repos-model.shareDat: data.key should exist')
+    const encodedKey = encoding.encode(data.key)
+    send('location:set', `?share=${encodedKey}`, done)
+  }
+
+  function cloneDat (state, data, send, done) {
+    assert.ok(Buffer.isBuffer(data) || typeof data === 'string', 'repos-model.cloneDat: data should be a buffer or a string')
```
juliangruber: this is taken care of by `dat-encoding`

models/repos.js
```diff
@@ -1,109 +1,205 @@
...
+
+    var key = encoding.decode(data).toString('hex')
```
juliangruber: can be refactored to `var key = encoding.toStr(data)`

models/repos.js
```diff
@@ -1,109 +1,205 @@
...
+    mkdirp(state.downloadsDir, function (err) {
+      if (err) return done(err)
+      var dir = path.join(state.downloadsDir, key)
+
+      mkdirp(dir, function (err) {
+        if (err) return done(err)
+
+        var opts = { key: key }
```
juliangruber: why not simply inline `opts`? it's small and not used elsewhere

package.json
```diff
@@ -7,45 +7,37 @@
   "description": "Dat Desktop App",
   "author": "Dat Team",
   "dependencies": {
-    "aliasify": "^2.1.0",
```
juliangruber: thanks for cleaning this up

models/repos.js
yoshuawuyts: still find it hard to read tho

models/repos.js
yoshuawuyts: oh nice!

models/repos.js
yoshuawuyts: not necessarily - if `undefined` is passed for example we'll get a weird error somewhere down in a package - this guards against that

models/repos.js
yoshuawuyts: ah yeah, nice one (:

models/repos.js
yoshuawuyts: oh yeah, I like tasks!

models/repos.js
yoshuawuyts: yup!

models/repos.js
yoshuawuyts: aye

lib/root-dir.js
yoshuawuyts: oh yeah sounds good - we should probably come up with a better way to handle options later, but re-adding :sparkles:

package.json
yoshuawuyts: :D

yoshuawuyts:
@juliangruber doneeeeee :sparkles:

models/repos.js
yoshuawuyts: (hope you ok with this? - I know I use assert a lot, but feel we're still at a spot where we regularly get weird errors that having lots of asserts can help us catch and trace bugs better)

models/repos.js
```diff
@@ -195,6 +195,14 @@ function createManager (multidat, onupdate) {
     var stats = dat.trackStats()
     dat.stats = stats
 
+    multidat.readManifest(dat, function (_, manifest) {
```
juliangruber: looks like this fn is only called once, however the metadata should update whenever the `dat.json` changes

elements/table.js
```diff
@@ -196,12 +194,15 @@ function tableElement (dats, send) {
 // ([obj], fn) -> html
 function createTable (dats, send) {
   return dats.map(dat => {
-    const progress = dat.stats.blocksTotal
-      ? Math.round(dat.stats.blocksProgress / dat.stats.blocksTotal * 100)
+    const stats = dat.stats && dat.stats.get()
```
juliangruber: shouldn't `dat.stats` always be set? also, this getter was more convenient to use: https://github.com/datproject/dat-desktop/pull/198/files#diff-66caf64acff533dbfc525b13489309cdL178

elements/table.js
```diff
@@ -196,12 +194,15 @@ function tableElement (dats, send) {
...
+    const progress = (stats)
```
juliangruber: why are you adding parens around single values? precedence over the identity is always true

lib/root-dir.js
juliangruber: the `mkdirSync` is missing still. why not just add back this file as a whole?

models/repos.js
```diff
@@ -1,116 +1,226 @@
...
-  const manager = new Manager()
-
-  ipc.on('link', (ev, url) => {
-    const key = encoding.decode(url)
-    manager.download(key, (err) => {
-      if (err) throw err
-    })
-  })
+function createModel () {
+  var manager = null
```
juliangruber: `let`

models/repos.js
```diff
@@ -1,116 +1,226 @@
...
-  // allow effects to have access to all state
-  function shareState (state, action, send, done) {
+  // boot multidat, create the ~/Downloads/dat directory
+  function startMultidat (send, done) {
+    var dbLocation = path.join(process.env.HOME)
+    var dbFile = path.join(dbLocation, '.dat-desktop-db.json')
+
+    var tasks = [
+      function (next) {
+        mkdirp(dbLocation, next)
```
juliangruber: `process.env.HOME` _always_ exists, no need to mkdirp it

models/repos.js
```diff
@@ -1,116 +1,226 @@
...
+      },
+      function (_, next) {
+        var db = toilet(dbFile)
```
juliangruber: const

models/repos.js
juliangruber: `dat-encoding` has sufficient guards now too, see https://github.com/juliangruber/dat-encoding/blob/master/index.js, but i'm also ok with leaving this here, for extra clarity

pages/main-delete.js
```diff
@@ -15,15 +15,18 @@ function view (state, prev, send) {
...
 
   return html`
     <body>
       ${sprite()}
       ${header}
       ${Table(archives, send)}
-      ${confirmModal(() => send('repos:deleteConfirm', link))}
+      ${confirmModal(() => send('repos:remove', {
+        confirmed: true,
+        key: link
```
juliangruber: having `key` _and_ `link` makes this confusing, do you think we can get it to just use one term across the codebase?

elements/table.js
yoshuawuyts: `const progress = stats` looks like an assignment without parens imo; always write my ternaries / arrow functions with parens; helps with readability

elements/table.js
yoshuawuyts: not a fan of getters / setters / proxies as an API b/c it hides what's going on; but yeah think you're right - it's set always now; was being somewhat defensive (:

models/repos.js
yoshuawuyts: good point

pages/main-delete.js
joehand: 👍 I'm guilty of this too in the CLI.

In my mind `key` the 64 character hash (mostly as a buffer), without any other characters. `link` is any think that contains a key (or references one), e.g. `dat://<key>`, `datproject.org/view/<key>`, `datproject.org/karissa/more-tweets-more-votes` (which resolves a key via API). Though I mostly refer to all of these as `key` in the CLI.

But we should decide on a [consistent terminology](https://docs.datproject.org/terms#dat-link).

elements/table.js
juliangruber: yeah, there's a balance with defensive programming and it's not easy to determine. however since we know there's always `dat.stats`, it's clearer this way. Otherwise you can read the code above and think otherwise.

elements/table.js
juliangruber: `const progress = (stats)` can be an assignment too though :P

juliangruber: My main point with style is just that it should be consistent over the whole project. Either stick to the existing style, or make the rest of the code base follow along in an atomic step.

yoshuawuyts:
just arrived at a place with internet so pushed final changes based on comments you made; tried pushing yest but somehow github wouldn't let me. Should be all good now I think - feel free to merge.

Big warning on that this might cause all sorts of bugs to pop up b/c we're handling more errors now; if you folks could go ahead and test things that'd be very valuable I reckon. Cheers!

juliangruber:
you force pushed over the commits i pushed to this branch, that's why it didn't let you

juliangruber:
you force pushed over commits i added to this branch, that's why github didn't let you

PR #205, opened 2017-02-10
fix dat boot screen

yoshuawuyts:
Makes it so the welcome screen is only shown when theres no dats, and stays dismissed after.


Kriesse:
approved, added changes

PR #220, opened 2017-02-11
allow recovery after critical crash

yoshuawuyts:
screen shot 2017-02-11 at 17 45 12

WIP. Closes https://github.com/datproject/dat-desktop/issues/213


karissa:
nice, is that supposed to be yellow instead of red?

good work!

elements/crash-modal.js
```diff
@@ -0,0 +1,83 @@
...
+
+  function render (onOk, onExit) {
+    return html`
+      <div class="relative flex flex-column justify-center ${prefix}">
+        <section class="pa4">
+          <h3 class="f4">Unexpected issue</h3>
+          <p class="mt3 mb4 f7 color-neutral-70">
+            There was an unexpected issue. We will need to close the app.
+            Sorry for the inconvenience and we will work to fix this ASAP.
+          </p>
```
karissa: something like 'There was an unexpected fatal issue. The app will now close. We have received a bug report and will work on it as soon as possible.'

karissa: 'You can try clearing the database to see if that will solve your issue, but will remove any settings. It will not delete your dat archives, though.'

Do we want to offer to clear all of the data? what does that mean? Dat data? I wonder if it would make sense to put a link to the GitHub issues instead of having a button to delete all of the dat data.

yoshuawuyts: Given our target audience is scientists it might make sense to keep them as far removed from GitHub as possible - remember before I knew how to code GitHub / sourceforge were v. scary places. Having concrete actions for them to perform by clicking a button might probably be the best we can do; bar making a "help" website, but I think that's out of scope for now.

Agree with you that we should mention the bug report and improve our copy more. Gonna update; keen to hear what you think.

yoshuawuyts:
Took forever to track this down, but fixed a bug that enables this functionality https://github.com/yoshuawuyts/nanocomponent/pull/6 - phewwwwwwwww

elements/crash-modal.js
```diff
@@ -0,0 +1,106 @@
...
+    `
+  }
+
+  function onexit () {
+    var window = remote.getCurrentWindow()
+    window.close()
+  }
+
+  function deleteData () {
+    const argv = minimist(remoteProcess.argv.slice(2))
```
juliangruber: we should refactor this to `lib/downloads-dir.js` again

elements/crash-modal.js
```diff
@@ -0,0 +1,106 @@
...
+    const downloadsDir = (argv.data)
+      ? argv.data
+      : path.join(app.getPath('downloads'), '/dat')
+    rimraf(downloadsDir, onexit)
+  }
+
+  function clearDatabase () {
+    const dbLocation = path.join(process.env.HOME)
```
juliangruber: that should also be in a lib, possibly together with `downloadsDir`, as it's used in other locations too

elements/crash-modal.js
yoshuawuyts: I think having it inline is fine for now; we probably want to shift away from CLI arguments anyway. Created an issue for architectural improvements down the road: https://github.com/datproject/dat-desktop/issues/223

elements/crash-modal.js
yoshuawuyts: Having it inline is fine; we should change this in the future anyway.

yoshuawuyts:
alright, merging!

Issue #223, opened 2017-02-13
architecture improvements for user accounts milestone


yoshuawuyts:
In preparation for user accounts we should probably take a closer look at how we're modeling our data. As per #152 we know our performance isn't great, and if we're going to add more complexity it might be worth taking a closer look at how we're doing things. I feel we're not doing a bad job, but we've got a few hacks right now that we should probably figure out to keep the project clean.

### determine when done initializing
We should have a way to know when all models are done initializing. Right now the `repos` model has some ready state in it, but there'll probably be more databases / files we'll be storing and need to wait on. It seems like having a single "done loading" indicator would be useful.

### preferences model
In [deskmux](https://github.com/yoshuawuyts/deskmux/tree/master/models) I've created a preferences model that holds all user configuration. This seems to work quite well. In our app we should probably merge the `main-view` model and all our minimist stuff; probably back it by a `toiletdb` instance. I feel having view-specific models seemed to not be a great idea. This would also serve for our stable / development versions to hold their own config rather than needing to rely on CLI arguments.

### too many modals
We have like 4 different variations of `main-view` + `modal` right now; there should be a better way I reckon. If we can do an unfancy way of passing down elements it'd probably be good; probably iterate on our `querystring-router` abstraction a lil.

### derived state
Deriving properties from other properties seems like a sane way to manage data. [twitter does it](https://medium.com/statuscode/dissecting-twitters-redux-store-d7280b62c6b1#.gf63gzzes) - I feel this is something we should probably think about / experiment with _outside the project_ to figure out how / if it will work for us. Hopefully this would allow us to reason about data easier.

## todos
- [ ] determine when done loading at an app level
- [ ] implement preferences model
- [ ] move all minimist stuff to preferences model
- [ ] move `main-view` model stuff to preferences model
- [ ] figure out how to reduce our modal count
- [ ] experiment with derived state

## See Also
- [twitter's redux architecture](https://medium.com/statuscode/dissecting-twitters-redux-store-d7280b62c6b1)

yoshuawuyts:
Another thing we should do is more testing. Core components that need testing right now:
- nanomorph (e.g. event copying and server rendering)
- nanocomponent (completely untested; have run into issues several times now)
- all our models

This should help us to refactor with more confidence rather than things breaking without us noticing. This probably also requires faster builds but that's a separate concern

yoshuawuyts:
@juliangruber and I are meeting up IRL this week; gonna chat about this

[...]

PR #225, opened 2017-02-13
update db location

yoshuawuyts:
In light of #223 it might make more sense to have a directory to store our config in; when moving our config from `indexedDB` into `toiletdb` this could serve to hold our config files. I suspect it might also help make migrations between versions easier. Thanks!


yoshuawuyts:
merging :sparkles:

karissa:
I have a commit in the integrations test branch that also allows us to pass a flag to customize the location.

PR #226, opened 2017-02-14
make sure that create new dat works as intended with stats

karissa:
This PR makes sure we are properly importing files into the dat and rendering the status. 

- We weren't calling it at all. Now it'll import the files and update the table on every file import event.
- Modularizes out the progress bar rendering code. 
- Removes the progress bar for local dats that the user owns (shared dats). This is consistent with the CLI. The reason we made this decision in the CLI is that it is very difficult to tell what the progress is for a dat now. So instead, it will say 'Watching for updates' where the progress bar used to be. 
- Because the dats could be downloaded remotely or created locally, the header says 'Status' instead of 'Downloaded'

### DIFF
--- a/elements/status.js
+++ b/elements/status.js
@@ -1,0 +1,101 @@
+const html = require('choo/html')
+const css = require('sheetify')
+
+const progressbar = css`
+  :host {
+    --progress-height: .5rem;
+    --bar-height: var(--progress-height);
+    --counter-width: 2.5rem;
+
+    min-width: 8rem;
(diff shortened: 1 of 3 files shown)


joehand:
The bit in `models/repo.js` looks good and consistent with CLI.

Sorry I had missed the no importing before =). That used to be done `automatically.`

karissa:
I see!

yoshuawuyts:
Nice!

PR #230, opened 2017-02-14
fix multiple deletions in a row

yoshuawuyts:
This fixes the issue where multiple deletions in a row caused the wrong `dat` to be referenced and crashed the app. The core issue at hand is most likely events not being copied over correctly and referencing old state. Thing seem to point to a `morphdom` issue. Made a note to test this in https://github.com/datproject/dat-desktop/issues/223#issuecomment-279685613.

In the meantime I've solved the issue by creating a temporary variable to store the pending key for deletion.

Had to move a lot of code around to pinpoint the issue; apologies for the rather large diff here.

### DIFF
--- a/elements/confirm-modal.js
+++ b/elements/confirm-modal.js
@@ -32,9 +32,10 @@ const prefix = css`
 module.exports = createWidget
 
 function createWidget () {
-  const modal = Modal({ render, onexit })
-
-  return modal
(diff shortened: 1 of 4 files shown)


yoshuawuyts:
@juliangruber thanks for the quick review! Merging :sparkles:

PR #236, opened 2017-02-15
fix delete 2 repos in a row

juliangruber:
Two parts were required to this:

- create a new `ConfirmModal` instance each time, because otherwise old instances would be recycled inproperly. It's not an issue perf wise to create a modal class every time for us fortunately
- use `send()` to set the uri back instead of `window.location.back`, which i think caused a race condition


juliangruber:
not fixed

juliangruber:
seems fixed!

juliangruber:
I fixed this by bypassing the routing layer for modals completely - which in my understanding isn't necessary anyway

elements/link-modal.js
```diff
@@ -120,12 +120,18 @@ function createModal () {
...
+  const modal = Modal({
+    onunload,
+    onexit,
+    render,
+    class: 'modal'
+  })
 
   return modal
 
   function onexit () {
```
juliangruber: can be refactored if we get a reference to the current element here, either via context or an argument:

```js
function onexit () {
  this.parentNode.removeChild(this)
}
```

or

```js
function onexit (el) {
  el.parentNode.removeChild(el)
}
```

I'd prefer the former

karissa: yeah the former makes sense

yoshuawuyts:
I've got some concerns by making UI elements in models, but happy to merge for now and I'll refactor it in a subsequent patch. Let's get this puppy out there :sparkles:

elements/link-modal.js
yoshuawuyts: Not a fan of using `this` in any form external facing API; feel it's best left for internals

PR #237, opened 2017-02-15
add network health

juliangruber:


karissa:
If we're going to say 'sick' next to a dat i'd like to explain a bit more about why its 'sick'.. maybe with a tooltip or question mark. Seems like we're introducing a new term here that many people might get confused/not understand why at first. And 'sick' sounds pretty strong as a word!

karissa:
Maybe something like 'fragile' or 'weak' would be better or maybe making the icon yellow or red when it's sick

juliangruber:
for now we're only using that as opinionated class names, could also go for something more neutral like `peers-0`, `peers-1`, `peers-many`.

karissa:
Oh great, no worries then -- for some reason I thought this was public to the user ^_^

Kriesse:
Good point though, let's opt for neutral class names. Will change it to Julian's suggestion.

PR #240, opened 2017-02-17
fix error messages

yoshuawuyts:
Had a lil regression with the error messages; this restores it (:


juliangruber:
are we getting string errors sometimes?

yoshuawuyts:
apparently hah; thought we didn't but yeah this fixes catches all cases I guess

Issue #206, opened 2017-02-10
Welcome screen should be dismissable with the keyboard


yoshuawuyts:
Caught myself pressing enter to dismiss it twice now; should probably allow for space bar / enter to dismiss it

PR #262, opened 2017-02-22
Makes welcome screen dismissible with Enter or Space

jekrb:
This pr addresses #206 (Welcome screen should be dismissible with the keyboard)

With these changes the user can press Space or Enter to dismiss the welcome screen.

At first I tried to add a keypress eventListener to the `<main>` element.  However, I learned that electron [requires a bit of machinery to listen to key events in the DOM](http://stackoverflow.com/questions/40763427/electron-does-not-listen-keydown-event).  I figured that it would be less code across the project if I added a keydown eventListener to the window, and then promptly removed it if the Enter or Space keys were pressed (also on the `onexit` method), and then send an action to the `toggleWelcomeScreen` reducer.

I hope these changes work well for you, and congratulations on the launch!

### DIFF
--- a/pages/main.js
+++ b/pages/main.js
@@ -89,17 +89,12 @@ function mainView (state, prev, send) {
       <div>
         ${sprite()}
         ${WelcomeScreen({
-          onexit: () => send('mainView:toggleWelcomeScreen', { toggle: false }),
-          onload: main => {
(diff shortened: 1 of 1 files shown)


karissa:
nice job!

Issue #266, opened 2017-02-23
ability to pause / resume downloads


juliangruber:
@Kriesse in https://github.com/datproject/dat-desktop/issues/184#issuecomment-281316401:

> As shown in the suggestion, I think we should tie this in with another feature: Users should be able to pause the transfer (download+upload) of a dat. In context:

![image](https://cloud.githubusercontent.com/assets/53684/23162905/83d1fa1c-f830-11e6-8022-202791ddbae1.png)

@yoshuawuyts please assign priority and milestone

juliangruber:
needs: paused state in multidat

juliangruber:
work has started here: https://github.com/datproject/dat-desktop/tree/add/toggle-pause

juliangruber:
needs two things:

1) multidrive: add {set,get}Meta                                                                                                                                        
2) dat-worker: add .{join,leave}{,Network}

juliangruber:
pull request here: https://github.com/datproject/dat-desktop/pull/291

PR #272, opened 2017-02-23
add new status icons. improve styling of network icon.

Kriesse:
This affects icons only and can be merged already.
The new icons are not yet added to the UI, but are needed for further styling work on #266.
Also contains a little improvement of the network status icon.


juliangruber:
The CI fails are on linux currently, but I'm working on that

Issue #278, opened 2017-02-27
linux travis builds failing


juliangruber:
This line eg should print detected electron versions: https://travis-ci.org/datproject/dat-desktop/jobs/205778457#L210

juliangruber:
better: deactivate for now until this is fixed

juliangruber:
inside the travis vm, this actually happens:

```bash
travis@6d4756e39f9c:~/dat-desktop$ ./node_modules/.bin/electron --version
travis@6d4756e39f9c:~/dat-desktop$
```

juliangruber:
fixed by https://github.com/datproject/dat-desktop/pull/303, linux builds are back!

karissa:
woot!

karissa:
can we close this then?

juliangruber:
yuss!

PR #279, opened 2017-02-27
travis: remove linux for now

juliangruber:
until travis linux builds are fixed, so I can focus on higher priority issues


mojoaxel:
will be fixed in #278

PR #308, opened 2017-03-20
Integrate archiver

karissa:
We need to think about the UI and can hide the UI for now until we figure it out. I reckon it'd be nice to merge this to force us to support this code-wise.

It changes `dbPaused` to be a metadata db in general for the dats we have on file, so if we decide to add other persistent options we don't end up having a ton of different toiletdb files.

Depends upon an update to toiletdb https://github.com/maxogden/toiletdb/pull/9/files

### DIFF
--- a/app.js
+++ b/app.js
@@ -32,6 +32,7 @@ persist(opts, (p) => {
   app.model(require('./models/main-view')())
   app.model(require('./models/window')())
   app.model(require('./models/repos')())
+  app.model(require('./models/archiver')())
   app.model(require('./models/error')())
 
   app.router([
--- a/elements/table.js
+++ b/elements/table.js
@@ -206,7 +206,6 @@ function row (dat, send) {
   }
 
   var historyButton = function () {
-    console.log(dat.historical)
     if (dat.historical === undefined) return html`` // loading
     return html`
       <input type="checkbox" checked="${dat.historical}" onclick=${toggleHistory}/>
--- a/elements/table.js
+++ b/elements/table.js
@@ -123,7 +123,8 @@ function tableElement (dats, send) {
             <th class="tl cell-3">Status</th>
             <th class="tr cell-4">Size</th>
             <th class="tl cell-5">Peers</th>
-            <th class="cell-6"></th>
+            <th class="tr cell-6">History</th>
+            <th class="cell-7"></th>
           </tr>
         </thead>
         <tbody>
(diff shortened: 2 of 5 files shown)


yoshuawuyts:
@karissa I think I'm missing the context here - why do you want to add this? I don't think I've seen any prior discussion about this on either IRC, Slack or previous issues - I'm a bit confused about the use case haha

karissa:
Hey @yoshuawuyts ! This is so that users who want to have the full history of a dat are able to keep all of the previous versions (archiver-server does that). This is something that is a definite feature we've talked about making available in the desktop app as well as a cli daemon. 

Future features will include also 'checking out' to a specific point in history and viewing diffs of history. We need to capture all of this in issues but its something we've talked about a bunch in user tests and in initial visions of the tool :) 

This is why maybe the UI shouldn't necessarily include anything at the moment until we have some designs, but we should be making sure that our infrastructure is supporting this kind of thing so when we do automatic updates its not a pain. In the process of implementing I found that the `dbPaused` was an issue. 

We could decouple that implementation from this PR though, if you think that would be a good idea.

joehand:
Sweet! I'd agree that we should think about storing history sooner rather than later, since if we don't ship with that now then the history disappears.

> This is so that users who want to have the full history of a dat are able to keep all of the previous versions (archiver-server does that). 

AFAIK setting the `storage` option on `createArchive` is all we need to store the full history. So we could store it in the dat folder instead of a home dir. I'm not 100% on how replication for the history works.

Though I'm not clear how this will work for hyperdrive v8 (it may always store history now).

> Future features will include also 'checking out' to a specific point in history and viewing diffs of history. 

The new hyperdrive stuff makes it super easy to [checkout history](https://github.com/mafintosh/hyperdrive/tree/v8#var-olddrive--drivecheckoutversion)! See [dat-yolo](https://gist.github.com/mafintosh/d1207355b1c66999c1fd3610dbb24110) as an example (though mafintosh may show us something cooler tomorrow).

karissa:
I think people will want to have the files *and* have copies of backups, and ideally those would be content addressed for diffing purposes (right?). Curious what the implications of hyperdrive v8 are, too, and excited about what @mafintosh is gonna show us

juliangruber:
While it's great to have the archiver server module ready and so easy to integrate, I would think a better final implementation is archived by starting with the requirements and UX design first, and then letting the code follow along.

What about we keep this in a branch for now which we keep rebasing onto master to make sure it is still well integrated?

In that respect I think it would be great to land the `dbPaused` -> `dbMeta` refactor into master beforehand.

yoshuawuyts:
> Though I'm not clear how this will work for hyperdrive v8 (it may always store history now).

As seen in yesterday's hacky hour demo hyperdrive 8 appears to have history built-in, so moving to the incoming upgrade should give us this for free I think

---

>  I would think a better final implementation is archived by starting with the requirements and UX design first

Yeah think that'd be the right approach for this; starting with design has been working rather well for our recent features - think it'd make sense to do that for this feature too

PR #312, opened 2017-03-22
elements/modal: fix missing button focus. closes #309

### LINKED ISSUE #309: add modal keyboard shortcuts
Currently when you try to positively dismiss a modal via pressing the enter key, something strange happens: Since the focus still is on the element you clicked then triggering the modal, another modal of the same kind is created and overlayed.

The solution I think is to set the focus on the modal's positive button.

juliangruber:
@yoshuawuyts do you maybe see a cleaner solution? Maybe something like

```js
button.green('Yes, Remove Dat', {
  class: 'fr ml3 confirm-button',
  onclick: ondelete,
  focus: true
})
```

### DIFF
--- a/elements/modal.js
+++ b/elements/modal.js
@@ -146,7 +146,10 @@ function confirmModal () {
 
     var deleteButton = button.green('Yes, Remove Dat', {
       class: 'fr ml3 confirm-button',
-      onclick: ondelete
+      onclick: ondelete,
+      onload: function () {
+        deleteButton.focus()
+      }
     })
 
     var exitButton = button('No, Cancel', {
@@ -255,7 +258,10 @@ function errorModal () {
   function render (message) {
     var exitButton = button.green('Ok', {
       class: 'fr ml3',
-      onclick: onexit
+      onclick: onexit,
+      onload: function () {
+        exitButton.focus()
+      }
     })
 
     return html`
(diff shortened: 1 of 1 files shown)


elements/modal.js
```diff
@@ -146,7 +146,10 @@ function confirmModal () {
 
     var deleteButton = button.green('Yes, Remove Dat', {
       class: 'fr ml3 confirm-button',
-      onclick: ondelete
+      onclick: ondelete,
+      onload: function () {
```
yoshuawuyts: I believe the first argument passed in is a reference to `el` so you could do:
```js
function (el) {
  el.focus()
}
```
(if I'm not mistaken)

yoshuawuyts:
Hmm, thinking about it this might conflict with the `onload` of the `modal` itself; nesting `onload` handlers doesn't work, and attempts at fixing it have stranded in the past. Maybs just to be sure it'd be worth moving the focus logic to the modal onload handler? Feel we might otherwise risk having an obscure error down the road haha 😅

juliangruber:
i hear yosh typing

yoshuawuyts:
Discussed in call and I'll change the PR to hook into the modal's `onload` event

yoshuawuyts:
this PR should be good to merge actually; found out we can nest `onload` calls just fine :sparkles:

PR #315, opened 2017-03-22
Choo v5

juliangruber:
A bunch happened here! I hope this isn't too much a pain to review:

- I changed choo 4 logic to choo 5 logic
- I refactored the dat manager logic out of the repos model. The dat manager shouldn't be coupled to choo in any way, so it helps in multiple respects to have it live in its own file. While doing that I also simplified it a little
- I renamed the `main-view` model to `welcome`, as it's only responsible for the welcome screen and thus the name is more obvious
- I renamed the `window` model to `drag-drop`, for the same reasons
- I removed `lib/param-router` as it wasn't used any more
- I renamed a bunch of events, to make them more human friendly. Mostly removing namespaces  like `error:quit` -> `quit` and making them more human like `repos:delete` -> `delete dat`. I would see reasons to undo this though as well.
- I removed cases where a model was using the event system to communicate with other parts of the same model, using functions instead
- While I was refactoring already, I moved the monkey patching required to run our integration tests to its own lib, to declutter the repos model
- I refactored `pages/main` by splitting out elements into `elements/empty` and `elements/welcome`. There's now no more css in this file :)
- In the end, all the QA passed, except known issues found and fixed in master after this branch was created

In total this is -90 lines of code, and a huge speed up (thanks to choo 5)! :)

Once the reviews are done,

- [x] I'll do the non trivial work of rebasing this branch onto master,
- [x] and update to use all `choo-*` dependencies from npm.


juliangruber:
btw, travis fails because `test:integration` is still active in this branch, which will be solved after rebasing onto master.

elements/empty.js
```diff
@@ -0,0 +1,83 @@
...
+            … or select one of your local
+            <br>
+            datasets and start sharing it.
+          </p>
+        </div>
+      </div>
+    </main>
+  `
+}
+
```
yoshuawuyts: I think standard might complain about extra newlines

juliangruber: do you mean in the future? currently it doesn't

models/welcome.js
```diff
@@ -0,0 +1,20 @@
+var xtend = Object.assign
+
+module.exports = welcomeModel
+
+function welcomeModel (state, bus) {
+  state.welcome = xtend({
+    show: false
+  }, state.welcome)
+
+  bus.on('repos loaded', function () {
```
yoshuawuyts: ohhh, we should probably decide on how to name events; in choo we usually delimit using `:` and camelcase for words. In the `dat` API we use whitespace to space words, and don't think we have namespaces. I feel we should like agree on something and then stick to it. Perhaps `:` to namespace and whitespace to split words?

package.json
```diff
@@ -8,9 +8,9 @@
   "author": "Dat Team",
   "dependencies": {
     "base-elements": "^2.2.1",
-    "choo": "^4.0.0-6",
-    "choo-log": "^3.0.1",
-    "choo-persist": "^2.0.0",
+    "choo": "github:yoshuawuyts/choo#v5",
+    "choo-log": "github:yoshuawuyts/choo-log#v5",
+    "choo-persist": "github:yoshuawuyts/choo-persist#v5",
```
yoshuawuyts: all versions are published now; install latest should work :sparkles:

models/welcome.js
yoshuawuyts: This is not a blocker btw, perhaps something to discuss either over text or face? feel we can figure it out in 5mins when talking haha

lib/monkeypatch.js
```diff
@@ -0,0 +1,8 @@
+if (process.env.RUNNING_IN_SPECTRON) {
```
yoshuawuyts: I'm not completely sure what this file does; could you maybs add a tiny comment for future reference? Thanks 🙏

elements/empty.js
yoshuawuyts: oh really? oops might be wrong then 😅 - nevermind :sparkles:

models/welcome.js
karissa: +1 on having a convention

lib/monkeypatch.js
karissa: This is for mocking the dialog

models/welcome.js
juliangruber: ok, created an issue here: https://github.com/datproject/dat-desktop/issues/318

Issue #185, opened 2017-01-27
let author add/edit metadata (title, author)


juliangruber:
We can use in place editing for setting `.title` and `.author` in `dat.json` via the desktop app!

yoshuawuyts:
@Kriesse assigned this to you as it'd be cool to have designs for this before working on it

Kriesse:
Cool! In future releases, we should also have a full edit view, as inline edits come with a few accessibility issues.
Here are mockups for inline-editing, almost the same except how we handle save/cancel:

# Version A (implicit Save, explicit Cancel):
Changes are saved when the user clicks out of the editable area. 
Changes are dismissed when the "x" is clicked.

### Edit and Save:
![inline-editing_a](https://cloud.githubusercontent.com/assets/53684/23515142/9b8acad6-ff69-11e6-8867-c683364149be.gif)

### Edit and Cancel:
![inline-editing_a_cancelled](https://cloud.githubusercontent.com/assets/53684/23515146/9d878860-ff69-11e6-91d6-5009c8ff8fe9.gif)

# Version B (explicit Save, implicit Cancel):
The edits are saved when the button "Save" is clicked. 
The edits are dismissed when the user clicks outside of the editable area.

### Edit and Save:
![inline-editing-b](https://cloud.githubusercontent.com/assets/53684/23515183/b4829a96-ff69-11e6-8a74-f99286d50691.gif)

### Edit and Cancel:
![inline-editing-b_cancelled](https://cloud.githubusercontent.com/assets/53684/23515193/b8723d78-ff69-11e6-8157-973e7eb156e4.gif)

juliangruber:
I agree with an eventual full edit screen, and for now I favor version B, as it's more explicit

karissa:
dependent upon having a dat.json file #255

yoshuawuyts:
We decided on the call to go for option B first, and see how well that works

yoshuawuyts:
@Kriesse it's been merged now, could you do the styling bits? Thanks!

Kriesse:
The behavior needs a but of fine-tuning.
- [ ] after click, show the current title in the editable field
- [ ] give an indication that it's saving and saved (see mockup)
- [ ] when clicking outside of the editable area, editing is canceled (right now only "escape" does that)

juliangruber:
since @yoshuawuyts is out for a week I'll help with this :)

juliangruber:
[...]

PR #329, opened 2017-04-09
Edit field

yoshuawuyts:
Implements the metadata edit field from #185. Still buggy, but getting there - mostly just need to knock out the reload bug and then we should be good. Think it's good to start reviewing. Thanks!

### todos
- [x] when restarting the app things can crash (needs investigating, related to the `dat.json` creation for sure)
- [ ] style this puppy
- [ ] prune unused deps


models/repos.js
```diff
@@ -144,6 +145,27 @@ function reposModel (state, bus) {
     document.body.appendChild(modal)
   })
 
+  bus.on('dats:edit-start', function (key) {
+    assert.equal(typeof key, 'string', 'dats:edit-start: key should be type string')
+    var dat = state.repos.values.find(function (dat) {
```
juliangruber: I think we talked about always using the key as a buffer internally and only using the hex encoding for external apis. Shall we go with this here too?

elements/table-title-field.js
```diff
@@ -0,0 +1,163 @@
...
+    emit = null
+    state = {
+      isEditing: false,
+      editTarget: null,
+      editValue: null,
+      isOwner: false,
+      title: null,
+      key: null
+    }
+  }
```
juliangruber: this function could be used for setting the initial state too

elements/table-title-field.js
```diff
@@ -0,0 +1,163 @@
...
+
+  function update () {
```
juliangruber: should we diff state here?

elements/table-title-field.js
```diff
@@ -0,0 +1,163 @@
...
+    var res = true
+    return res
+  }
+
+  function render (newDat, newState, newEmit) {
+    if (newEmit) emit = newEmit
```
juliangruber: does the emit function actually change between invocations of `render()`? otherwise we could pass it to the component constructor instead

elements/table-title-field.js
```diff
@@ -0,0 +1,163 @@
...
+    if (newDat) {
```
juliangruber: to me this variable would be cleared just called `dat`, as I haven't seen this naming pattern with components before

elements/table-title-field.js
```diff
@@ -0,0 +1,163 @@
...
+      state.isOwner = newDat.isOwner
+      state.key = newDat.key.toString('hex')
+      state.title = newDat.metadata.title || '#' + state.key
+    }
+
+    if (state.isEditing) return component.emit('render:active')
+    else return component.emit('render:inactive')
```
juliangruber: how does this work? does your event emitter actually return something from `.emit()`?

elements/table-title-field.js
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  function renderInactive () {
+    state.editValue = ''
```
juliangruber: should a render function set state?

juliangruber: In general this pattern of managing a component's own state seems confusing still...I'll see how I think when I spend more time with this pattern

elements/table-title-field.js
```diff
@@ -0,0 +1,163 @@
...
+
+  function renderActive () {
+    if (!state.isEditing) {
+      state.isEditing = true
+      attachListener()
+    }
+
+    setTimeout(function () {
+      self._element.querySelector('input').focus()
+    }, 0)
```
juliangruber: this looks leaky, as it's making an assumption about when the dom is going to be mounted. Could we instead use `onload="..."` for this?

elements/table-title-field.js
yoshuawuyts: yeah, been thinking about this - perhaps we should :sparkles:

elements/table-title-field.js
yoshuawuyts: yeah. Wasn't sure what the cleanest pattern here was - know this is not in line with Node's event emitters. We could use a regular function instead, but then we use all references tho `this` and cool stuff like logging. Think maybe the `.on()` keyword is the confusing part here. Have any suggestions on how to make this better?

elements/table-title-field.js
yoshuawuyts: Yeah, I feel you - not stoked about some of these things either. Although in a way having isolated, component-specific state doesn't feel like the worst. I'm not sure thoo

elements/table-title-field.js
yoshuawuyts: `on-load` wouldn't work here b/c we can't nest it at the moment. With you tho, I wish it worked as I'm not a fan of this pattern either. Looked into it last night and think we just need to debug `on-load`

elements/table-title-field.js
yoshuawuyts: ooops, yeah we should :sparkles:

elements/table-title-field.js
juliangruber: i like what react does here, having `this.state` and `this.props` separate. `.props` is what the component is called with `.state` is the internal state it manages.

elements/table-title-field.js
juliangruber: ah, yeah that pattern definitely is confusing, also confused me looking at how `component.on('render:active', renderActive)` would work. Thinking, why would it make sense to `return` something from inside an event handler? Someone looking at the code from the outside has no reason to believe that this is not using node's EventEmitter API.

At least I would expect the methods to be called differently, like `component.handle('render:inactive')` maybe?

elements/table-title-field.js
juliangruber: I'm working on a PR to nanocomponent to add this stuff

juliangruber: https://github.com/yoshuawuyts/nanocomponent/pull/25

elements/table-title-field.js
juliangruber: Woah yes, if we can get `on-load` to support that, that would be really awesome! I couldn't find an issue in the `on-load` repo though talking at this, can you maybe point me to some prior discussion?

yoshuawuyts: It's tricky tho https://github.com/shama/on-load/pull/19

yoshuawuyts: Also created https://github.com/shama/on-load/issues/21

yoshuawuyts:
@juliangruber integrated all feedback that's possible for now - gonna leave some stuff like the `on-load` things on the backburner; not sure how long it'd take to solve and don't think we should block it on that. Think also once we get the new `microcomponent` API out we can migrate in a separate patch.

@Kriesse think it's ready for styling now! If you don't have time to pick this up right now maybs it's an idea to merge this patch and do styling in a separate patch (if @juliangruber signs off on this); think this is blocking development slightly so the faster we can move it the better I reckon.

Thanks! :sparkles:

yoshuawuyts:
Woot! :D

PR #336, opened 2017-04-20
s/repos/dats

yoshuawuyts:
We were referring to it as "dats" everywhere anyway,
figured we might as well rename our code.

This shouldn't cause any regression b/c the dats aren't stored in a set file anyway, this is internal only. Thanks!


juliangruber:
awesome!

juliangruber:
`lol-repos` :D

PR #338, opened 2017-04-20
appease the ci gods

yoshuawuyts:
Dep CI isn't happy, even if semver kicks in. This should fix it. Thanks!


juliangruber:
choo-expose and nanologger are unlicensed, that's why it complains it seems

yoshuawuyts:
Yeah so I fixed that but apparently it doesn't pick up the semver patch
change :(

juliangruber:
you can re-run the dependency-ci builds through the web ui, i just triggered that so it should be fine soon

yoshuawuyts:
hmm, apparently it's still not working. I think we ought to just disable dependency-ci then :/

screen shot 2017-04-21 at 13 14 00

yoshuawuyts:
blep, just disabled dep ci - can't get it to pass so if it's gonna get stuck on a false positive I don't think it's worth having around

juliangruber:
yeah, dependency-ci seems to be behind!

Issue #341, opened 2017-04-27
styling of failed dats


juliangruber:
![screen shot 2017-04-27 at 15 40 33](https://cloud.githubusercontent.com/assets/10247/25485930/e54054b2-2b5f-11e7-913c-e3b41942bdc8.png)

While we display the error message when a dat fails to load - for example because it is already being served via the cli - the styling should be improved.

Kriesse:
@julian Looks like a styling issue that'll come up again for other errors on a dat, how can we make it so that it's easy to re-use (and reproduce for me locally)?

Kriesse:
@yoshuawuyts As per discussion, let's get this fixed – can you point me to where to get started with styling?

juliangruber:
@Kriesse one way to reproduce is to create a dat in desktop, quit desktop, share the dat via cli, and while still sharing open desktop again

juliangruber:
@Kriesse I think in `elements/table-row.js` you need to look at `errorRow()` for the DOM that's generated and add a css block for this as well.

Kriesse:
> one way to reproduce is to create a dat in desktop, quit desktop, share the dat via cli, and while still sharing open desktop again

@juliangruber ha, just tried that, but that only gives me the error in the CLI :)
any ideas?

juliangruber:
oh, what error are you getting in the cli?

juliangruber:
was desktop maybe still running while you started the cli?

juliangruber:
this is done in master :)

PR #346, opened 2017-05-02
style failed dat error #341

Kriesse:


elements/table-row.js
```diff
@@ -323,20 +323,25 @@ function HexContent () {
...
-      <td class="tr cell-4 size">
-      </td>
-      <td class="cell-5">
+      <td class="cell-2" colspan="4">
+        <div class="cell-truncate color-red">
+          <h2 class="f6 normal">
+            Error
+          </h2>
+          <p class="f7">
+            Error initializing dat in ${err.dir}
```
karissa: Can we say something less techy, like 'Could not share ${err.dir}'

Kriesse: 👍

juliangruber:
The build seems to fail because some of the copy was verified in the integration tests:

```js
 t.ok(val.indexOf('create new dat') > -1, 'has create new dat text')
```

You can fix those in `tests/welcome.js` and then verify locally by running `npm run test:integration`

karissa:
ok i'll fix that

PR #348, opened 2017-05-02
BUG: Only owner should be able to inline edit the title.

### LINKED ISSUE #347: Don't allow renaming of read-only Dats
Via inline-editing, I can rename a read-only dat (=edit dat.json). 
That probably shouldn't happen.

karissa:
Only owner should be able to show the inline file editing interface. fixes #347


juliangruber:
Mergarita!

PR #350, opened 2017-05-02
Change copy to be consistent with CLI

karissa:
screen shot 2017-05-02 at 2 13 29 pm

New copy should improve conversion rate of users that go from downloading the binary to sharing data!

CLI uses 'clone' for a Dat -- open to replacing 'download' with that, except right now there is no multi-writer so 'download' is more accurate to what is actually happening here.

Please discuss, thanks! 🎉


clkao:
somehow travis is green on this PR but breaks master after the merge.  It seems to be caused by "create new dat" is being replaced by "share a folder" and alike.

karissa:
Fixed on master tahnks cl

PR #358, opened 2017-05-15
throttle stats updates to once per second. closes #357

### LINKED ISSUE #357: debounce speed updates
no need to rerender the network speed at 60fps

juliangruber:
The speed display was updating so fast that it was hard to use, this limits it to one update per second.

In another iteration we can make sure that all the rows update at the same time, in synchronicity.

This also reduces CPU load :)


Kriesse:
@juliangruber Commits look good, but not sure how to visually test?

juliangruber:
when you download a dat, the network speed should only update once a second instead of way way more

PR #382, opened 2017-05-19
add ESC to dismiss intro

juliangruber:
for some reason the onload handler isn't firing. @yoshuawuyts do you have an idea?


yoshuawuyts:
I think we should turn this into a component. Having onload as an attribute was always flaky; hence we removed it in `bel@5` (https://github.com/shama/bel/pull/76)

juliangruber:
ah, ok! it is a component already, so i'll add the `.on('load')` handler

juliangruber:
all fixed up and working!

Issue #271, opened 2017-02-23
link to issue tracker from within app


juliangruber:
Do we for example want to link to our github project from the welcome screen? Just thinking about whether people do know where to submit issues.

Also, maybe add this to the Help menu item and the error screens.

- [x] add a menu trigger to header that toggles a dropdown
- [x] style it
- [x] menu toggles when you click the icon again (right now it doesn't go away)
- [x] link to issue tracker opens in browser
- [x] version number is dynamic (if possible)

yoshuawuyts:
Think it would indeed be useful to have a view with more info about how to provide feedback / link to IRC, twitter, github and the like. Assigning to @Kriesse because we probably need some design for this first

Kriesse:
I think we should have a menu trigger (e.g. a little "i" icon) in the top right of the nav bar. on click, we show a drop down menu. Same style as in datproject.org.
We can place version number, link to github (and later any profile-related links) there.

## menu dropdown in datproject.org:
![image](https://cloud.githubusercontent.com/assets/53684/26343061/d94ff32a-3f9a-11e7-89b1-d3276c013e98.png)

Kriesse:
@juliangruber @yoshuawuyts Can one of you help with this?
> add a menu trigger to header that toggles a dropdown

I'll do the styling :)

Kriesse:
Can I get the menu trigger please? 🙏

juliangruber:
working on it!

juliangruber:
Pull request here: https://github.com/datproject/dat-desktop/pull/406

Kriesse:
Initial styling done. @juliangruber I need help with some details, see list above.

![image](https://cloud.githubusercontent.com/assets/53684/26577236/094e1dea-452c-11e7-9ec7-32b8f21c8dc9.png)

juliangruber:
@Kriesse all issues addressed!

PR #406, opened 2017-05-30
add basic menu to header

juliangruber:
for #271 

The menu is still unstyled and the button to open the menu isn't clickable yet (css issue I think).

Once we have all the links in there we need, I'll make them open the target in a browser, as expected.


Kriesse:
@juliangruber What's missing here? Tested visually, LGTM!
image

PR #438, opened 2017-08-18
travis: test on node 8

ralphtheninja:
npm bundles with node 8 aborts installation if the browserify step fails, see https://github.com/datproject/dat-desktop/issues/436


ralphtheninja:
Interesting. It builds just fine on node 8 o.O, but not on my mahine *rips hair*

Issue #434, opened 2017-07-28
What about a daemon?


soyuka:
Hi there!

I'm currently hacking around with dat and my goal is to use dat on distant servers rather then at home.

Today, there are 2 user interfaces to manage dats:

1. `dat-cli` 
2. `dat-desktop` (uses `multidat`)

I want to be able to manage lists, and it felt natural to build a `daemon` [poc here](https://github.com/soyuka/dat-daemon). 

It got me thinking into having a distant web interface. Dat-desktop is good and has everything one may need. 

Would you be interested in a patch that separates the `client` (html, css, front js) from `electron`?

To ease further usage, do you think that a daemon would be a viable option for `dat-desktop`'s "server" side so that it could be used anywhere? The `daemon` would be responsible of managing `dats` elements (add, remove, list, stat). 

Everything could go through the daemon (for electron `net.connect` + `dat-web` - maybe fork the daemon, possibility to share with a closed app, quick-menu etc.).

karissa:
That sounds like a great idea.

soyuka:
I see that you run the tests on node 7.0, though it's neither LTS nor Current.
Do you wish to support LTS? What about moving to Current so that we can use `async/await`?

To start working on this, IMHO, the code responsible to manage the *dat-list* should be a module.

I've two suggestions for what could be done:

1. don't use any tcp from electron, rather directly use the *dat-list* module with the ipc bus
2. use a daemon from electron with a bridge ipc => tcp

In both case, having this *dat-list* module this would ease the implementation of a daemon. 

WDYT?

karissa:
Yeah, love this dat-list idea!

soyuka:
> Do you wish to support LTS? What about moving to Current so that we can use async/await?

:stuck_out_tongue:

joehand:
> Do you wish to support LTS? What about moving to Current so that we can use async/await?

Ya, in general we target anything covered in Node's LTS or maintenance (node 4, 6, 8). Moving test targets from 7 -> 8 sounds good for this project and we've tried to do that in general, just a bit behind here. Because this package gets bundled and distributed, it can use anything. But any modules that will be used in other projects should be able to be used in older node versions. This would likely apply, for example, to the daemon.

You're free to use new js features but be warned that some of use are still on node 6 and don't really use async/await yet, so it may impact contributions and and PR reviews. Optimally, you'd provide a way to build with older node versions for development.

[...]

PR #440, opened 2017-09-04
 (WIP) Implement daemon

soyuka:
This is really a work in progress!

-------

I need to be able to access the `manager` that has been created in `models/dats`. 

I have a few solutions, maybe you can help me find the best (anyway it helps me to write thoughts 😄):

1. use the state to store the manager (I don't like this because of many issues especially that the state doesn't need this complex object, and that the manager is created asynchronously)
2. use the bus to get the manager `bus.once('dats:manager', () => {})` + `bus.emit('dats:manager', manager)` once created
3. try to export the whole manager thing to a new file, but as it's initialization is asynchronous it'll be hackish to share the resolved manager
for example:
```
var manager
module.exports = getManager

function getManager() {
 if(manager) { return Promise.resolve(manager) }

return createManager()
.then(manager => manager)
}
```

WDYT?

 #434

### DIFF
--- a/app.js
+++ b/app.js
@@ -4,6 +4,15 @@ const log = require('choo-log')
 const css = require('sheetify')
 const choo = require('choo')
 const xtend = Object.assign
+const remoteProcess = require('electron').remote.process
+const remoteApp = require('electron').remote.app
+const minimist = require('minimist')
+const path = require('path')
+
+const argv = minimist(remoteProcess.argv.slice(2))
+const downloadsDir = (argv.data)
+  ? argv.data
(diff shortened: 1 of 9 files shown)


karissa:
Hey @soyuka this looks really cool. What do you think about putting initialization into a new file? Even though it's hacky, you only need to initialize it once so maybe it's not that huge of a deal right now.  What else is there still to do here?

soyuka:
I'm closing this in favor of a small daemon I've come up with  in https://github.com/soyuka/dat-daemon 

We could totally hook up the desktop/cli app with such a daemon not sure it's worth it though. Feel free to ping me if you want me to do this anyway!

PR #442, opened 2017-10-12
turn watch on by default

joehand:
This default changed in dat-node v3 but may haven't called it out.


juliangruber:
I've looked at this again but tests are still failing. Does anyone know what's going on here?

PR #444, opened 2017-11-25
fix $HOME on windows. closes #443

### LINKED ISSUE #443: path.join in dats.js causes problems with windows
Hi folks,

TLDR; Are there any plans to support windows at this state of the project?

-------------------

Just tried to compile and run the project on windows 10. .... Many manuals steps needed, since bash build scripts fail, atm. .... 

System seems to fail at startup already. ....

It seems this line https://github.com/datproject/dat-desktop/blob/f8c71b4c974f2c8f2e0e0b9c778f747c830486dd/models/dats.js#L44

causes a problem with an `undefined` path ... 

Solution seems to be something like this: https://stackoverflow.com/questions/9080085/node-js-find-home-directory-in-platform-agnostic-way

juliangruber:


karissa:
nice lgtm

PR #459, opened 2018-01-28
fixes typo and delete dat functionality + migrate buttons

aks-:


app/components/table-row.js
```diff
@@ -175,10 +176,10 @@ const Row = ({ dat, shareDat }) => (
       <IconContainer className="flex justify-end">
         <FinderButton dat={dat} />
         <LinkButton onClick={() => shareDat(`dat://${dat.key}`)} />
-        <DeleteButton />
+        <DeleteButton onDeleteDat={() => onDeleteDat(dat.key)}/>
```
juliangruber: can we refactor this to be

```js
<DeleteButton onClick={() => onDeleteDat(dat.key)}>
```

this way we don't need to touch the `DeleteButton` component

app/actions/index.js
```diff
@@ -91,3 +91,12 @@ export const addDat = key => dispatch => {
     }, 1000)
   })
 }
+
+export const deleteDat = key => dispatch => {
+  dispatch({ type: 'REMOVE_DAT', key })
+  const path = `${homedir()}/Downloads/${key}`
+  Dat(path, { key }, (error, dat) => {
```
juliangruber: since the `dat` object has been created already, we could optimize this by storing open dats in a map above:

```js
const dats = new Map()
// ...
dats.set(dat.key, dat)
```

and then here:

```js
const dat = dats.get(key)
dat.close()
dats.delete(key)
```

app/reducers/index.js
```diff
@@ -43,6 +43,11 @@ const redatApp = (state = defaultState, action) => {
           }
         }
       }
+    case 'REMOVE_DAT':
+      let { [action.key]: del, ...remainingDats } = state.dats
+      return {...state,
+        dats: remainingDats
+      }
```
juliangruber: smart! This can be simplified to:

```js
case 'REMOVE_DAT':
  const { [action.key], ...dats } = state.dats
  return {...state, dats}
```

juliangruber:
Thank you for taking a lead on this! :)

juliangruber:
Created an issue for this: https://github.com/datproject/dat-desktop/issues/464

app/actions/index.js
aks-: @juliangruber  implemented, thanks for tips, really appreciate It ^_^

app/components/table-row.js
aks-: I made it work like LinkButton

PR #480, opened 2018-01-29
Add standard

juliangruber:
It's definitely a weird choice of standard to use single quotes for JSX attributes, we might change that with an option if others agree. Otherwise, standard!

This will help us maintain consistent code style and best practices across all contributors.


martinheidegger:
> It's definitely a weird choice of standard to use single quotes for JSX attributes

Weird, yes, but i kinda like it 😅 

It needs a rebase, else LGTM

martinheidegger:
Travis seems to fail because the package-lock.json doesn't contain standard.

juliangruber:
I'm adding it, my local npm just broke :O

PR #488, opened 2018-01-30
contributing: add pull request time frame

juliangruber:


juliangruber:
This is the other way around: If a main contributor creates a pull request and doesn't receive review in this time frame, it should be good to merge. A pull request from a non main contributor should never be merged without prior review from a main contributor.

juliangruber:
Maybe you had a different goal then?

juliangruber:
what I wrote applies to both external and internal contributors:

- for an external contributor it means not to worry if their PR doesn't receive any attention for 36h
- for an internal contributor it means to wait for 36h before deciding there won't be more reviews and then deciding on whether to merge or not

martinheidegger:
You are right: there are two goals. What I was focussing on was two things:

1. Weekends. People need to know that other people rest ^_^
2. To Avoid confusion if a PR is merged as is without a comment.

of course your goal would be completely lost in my suggestion 😞

juliangruber:
how about this? :)

Issue #472, opened 2018-01-28
Drag and Drop


juliangruber:
- [X] drop on window https://github.com/datproject/dat-desktop/blob/master/models/drag-drop.js -> https://github.com/datproject/dat-desktop/pull/491

- [x] drop on app icon https://github.com/datproject/dat-desktop/blob/master/index.js#L58

PR #491, opened 2018-01-30
add dat drag/drop. #472

juliangruber:
This lets you share a folder by dropping it on the window


martinheidegger:
Added #497 and #496 after reviewing this PR

PR #493, opened 2018-01-30
Add inspect component

aks-:


app/reducers/index.js
```diff
@@ -7,6 +7,10 @@ const defaultState = {
       link: null,
       copied: false
     }
+  },
+  inspect: {
+    key: null,
```
juliangruber: can this be simplified into one variable? just `inspect: { key: null }` or `inspect: { key: ... }`

app/reducers/index.js
```diff
@@ -46,6 +50,28 @@ const redatApp = (state = defaultState, action) => {
...
+        inspect: { ...state.inspect,
+          key: null,
+          show: false
+        }
+      }
+    case 'ADD_FILE_TO_DAT':
+      return { ...state,
+        dats: { ...state.dats,
+          [action.key]: {...state.dats[action.key],
+            files: action.files
```
juliangruber: wording is misleading here: the event name suggests a single file is added, however looking at the code it's overwriting the whole files array. to be consistent with other events, it should be renamed to `DAT_FILES`

app/components/table.js
```diff
@@ -39,8 +39,8 @@ const StyledTable = styled.table`
   }
 `
 
-const Table = ({ dats, shareDat, onDeleteDat }) => (
-  <main>
+const Table = ({ dats, showInspect, shareDat, onDeleteDat, inspectDat }) => (
+  <main style={{ display: showInspect ? 'none' : 'block' }}>
```
juliangruber: I think the cleanest solution here is not to `display:none` but not render the component at all. What about renaming the `showInspect` property to simply `show` and returning an empty `<Fragment />` in that case?

So it would be `<Table show={false}>` if inspect is to be shown.

app/components/inspect.js
```diff
@@ -0,0 +1,140 @@
...
+`
+
+const Label = ({ children }) => (
+  <LabelStyle className="mb2">
+    {children}
+  </LabelStyle>
+)
+
+const Inspect = ({ dat, closeInspectDat, addFilesToDat }) => {
+  if (!dat) return <div style={{ display: 'none' }}></div>
```
juliangruber: similar to the comment on the `table` component, just return an empty `<Fragment />`

app/components/inspect.js
```diff
@@ -0,0 +1,140 @@
...
+  const peers = dat
+    ? dat.peers
+    : '…'
+
+  return (
+    <main className="flex flex-column">
+      <div className="flex flex-column flex-auto">
+        <DetailHeader className="flex items-center">
+          <div className="w3">
+            {icon('hexagon-down', {className: 'w2 center color-neutral-30'})}
```
juliangruber: should be

```js
<Icon name="hexagon-down" className="w2 center color-neutral-30" />
```

app/components/file-list.js
```diff
@@ -0,0 +1,58 @@
...
+`
+
+const List = ({ dat }) => (
+  <FileListContainer>
+    {dat && dat.files && dat.files.length
+        ?
+        <FileListTable className="w-100 f7 f6-l ">
+          <tbody>
+            {dat.files.map(file => {
+              var size = (Number(file.size) == file.size) && file.isFile
```
juliangruber: should be `const size` (es6 ftw 😃 )

app/components/file-list.js
```diff
@@ -0,0 +1,58 @@
...
+                ? bytes(file.size)
+                : ''
+              return <tr key={file.path}>
```
juliangruber: this reads nicer as

```js
return (
  <tr key...
)
```

app/components/file-list.js
```diff
@@ -0,0 +1,58 @@
...
+    tr:odd td {
+      background-color: var(--color-neutral-04);
+    }
+  }
+`
+
+const List = ({ dat }) => (
+  <FileListContainer>
+    {dat && dat.files && dat.files.length
+        ?
```
juliangruber: also use parens here:

```js
dat && dat.files && dat.files.length
  ? (
      <FileList> ...
    )
  : (
      <div> ...
    )
```

app/actions/index.js
```diff
@@ -4,6 +4,7 @@ import Dat from 'dat-node'
 import { encode } from 'dat-encoding'
 import { homedir } from 'os'
 import { clipboard } from 'electron'
+import mirror from 'mirror-folder'
```
juliangruber: this is missing from `package.json`

aks-:
@juliangruber  should I rebase or merge here?

martinheidegger:
@aks- usually rebase.

juliangruber:
works like a charm, thank you so much!

PR #500, opened 2018-02-04
add persist state on disk. closes #467

### LINKED ISSUE #467: Persist state between app restarts
PR: https://github.com/dat-land/dat-desktop/pull/500

juliangruber:


app/actions/index.js
```diff
@@ -191,3 +203,12 @@ export const dropFolder = folder => async dispatch => {
   if (!isDirectory) return
   addDat({ path: folder.path })(dispatch)
 }
+
+export const loadFromLocalStorage = () => dispatch => {
+  const blob = localStorage.getItem('state')
+  if (!blob) return
+  const dats = JSON.parse(blob)
+  for (const dat of dats) {
+    addDat(dat)(dispatch)
```
martinheidegger: adding to my former point: This should be called something other than dat  in order to have structure <-> name consistency. `datPath`?

app/actions/index.js
```diff
@@ -191,3 +203,12 @@ export const dropFolder = folder => async dispatch => {
   if (!isDirectory) return
   addDat({ path: folder.path })(dispatch)
 }
+
+export const loadFromLocalStorage = () => dispatch => {
+  const blob = localStorage.getItem('state')
+  if (!blob) return
+  const dats = JSON.parse(blob)
+  for (const dat of dats) {
```
martinheidegger: adding to my former point: This should be called something other than dat  in order to have structure <-> name consistency. `datPath`?

aks-: Hm,  I think `dat` is fine given `key` and `path` are properties on dat?

martinheidegger: `dat` is an object created with `dat-node` - no?

juliangruber: I agree, renamed it to `datOpt`

juliangruber:
Adding to the point that it's not compatible with previews Dat Desktop yet: Pause/Resume also isn't persisted yet

juliangruber:
@martinheidegger mind giving this another review?

martinheidegger:
@juliangruber I think you saw the points I wanted to fix. I don't have the time to prepare another PR soonish. Do you want to pick some changes and apply them or keep this PR as is?

juliangruber:
tbqh, this seems to be working good enough and can always be refactored later on (we now know the points, as you said). for our current deadline i'd way rather focus on features like the viewers.

PR #501, opened 2018-02-04
Make table title field editable

aks-:


app/components/table-row.js
```diff
@@ -161,15 +162,17 @@ const DeleteButton = ({ ...props }) => (
...
+const Row = ({
+  dat,
+  editing,
+  shareDat,
+  onDeleteDat,
+  inspectDat,
+  updateTitle,
+  makeEditable,
+  editTitle,
+  deactivate
```
aks-: I don't like adding above props, any suggestions on how to refactor this?

app/components/title-field.js
```diff
@@ -0,0 +1,141 @@
...
+        <div>
+          <Overlay onClick={() => this.deactivate()} />
+          <EditableField className='bg-white nt1 nb1 nl1 pl1 shadow-1 flex justify-between'>
+            <input
+              className='bn f6 normal w-100'
+              defaultValue={editValue || title}
+              onKeyUp={ev => this.handleKeypress(ev)}
+              ref={input => {
+                this.titleInput = input
+              }}
```
aks-: Can i do it without having to use `ref`?

martinheidegger:
Why did you implement this yourself, rather than using an existing component such as https://github.com/kaivi/riek ?

aks-:
@martinheidegger Hm, that's good point.... I will go through that library and ones like that to see if they would be good fit and refactor it with them

aks-:
@martinheidegger I looked into it and libraries like this one for react. This one seem to be most mature, but there are certain problems with it when it comes to our usage. There is no way to tell if RIEInput is selected or clicked upon, hence we can't tell if we are already in editing mode and we should show the `save` button. Also this project doesn't look like maintained well. I saw a lot of PR(s) unanswered if we make a pr into that for the feature. For instance take a look at this one - https://github.com/kaivi/riek/pull/65#issuecomment-354421879

I think we should first get it working and write our own library to replace the refs eventually

martinheidegger:
If it doesn't work, it doesn't work - thank you for giving it a try: It seems like `riak` needs a new maintainer. That being said: I think this needs a rebase. I found a few issues before that I didn't want to mention before a rewrite to riak - would be nice to add them to a rebased version.

aks-:
Yes I also found few, I am closing it and push a branch and make a pr again. Still working on it....

martinheidegger:
_Note: Its okay to force-push to a PR branch :) _

aks-:
Right :D but I thought I would push it upstream and make a pr from there. Also I didn't know about the pattern we used for table row refactor, now I can use that to make this one more clearer.. and some relatively better naming. So i thought it would be okay to make entirely separate PR 🤔

PR #502, opened 2018-02-05
Add intro screen

juliangruber:
New version of the original PR https://github.com/dat-land/dat-desktop/pull/498 by @aks-, with the branch available in this repository.

### DIFF
--- a/app/actions/index.js
+++ b/app/actions/index.js
@@ -175,5 +175,5 @@ export const dropFolder = folder => async dispatch => {
 }
 
 export const openHomepage = () => shell.openExternal('https://datproject.org/')
-export const nextIntro = (screen) => ({ type: 'NEXT_INTRO', screen })
+export const nextIntro = screen => ({ type: 'NEXT_INTRO', screen })
 export const hideIntro = () => ({ type: 'HIDE_INTRO' })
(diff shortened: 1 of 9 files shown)


juliangruber:
LGTM now :)

aks-:
yeap, looks good and works good now

juliangruber:
> I merged #507 into this branch (to test it out) and it shows the tutorial screen for a little until the persisted dat data is loaded. This can be partly mitigated by having a dedicated showIntro action rather than a hideIntro action. So my request would be to flip that the default-state is not intro but rather loading or startup with a dedicated action triggering the intro.

This can also be reversed: If we do that and the user opens the app when the intro needs to be shown, there will be a quick flash too.

I think this means we need to introduce not just `showIntro` but also `showTable`, so there's an anticipated state before either are visible.

martinheidegger:
>> So my request would be to flip that the default-state is not intro but rather loading or startup with a dedicated action triggering the intro.

Well, the "loading" screen in my opinion should be a show/hide of the browser window:

https://github.com/electron/electron/blob/master/docs/api/browser-window.md#using-ready-to-show-event

juliangruber:
Do you mean we shouldn't show the window at all before it has finished loading its internal state?

martinheidegger:
@juliangruber exactly.

juliangruber:
Let's fix the flashing in another PR, ok?

juliangruber:
btw, I can't approve because I created this PR (using @aks-'s branch), so someone else will have to do so.

PR #503, opened 2018-02-05
add protocol handler. closes #473

### LINKED ISSUE #473: Protocol handlers
PR: https://github.com/dat-land/dat-desktop/pull/503

juliangruber:
#473


martinheidegger:
Somehow I fail to see how this works, wouldn't it need to actually register protocols?
https://electronjs.org/docs/api/protocol

juliangruber:
Those protocols are for when you need to register a custom one to be used inside your electron app, however not to interface with different apps.

See also

- https://github.com/dat-land/dat-desktop/blob/react/package.json#L67
- https://www.electron.build/configuration/configuration

PR #506, opened 2018-02-06
optimize and refactor table-row

aks-:


app/components/table.js
```diff
@@ -73,14 +73,7 @@ const Table = ({
...
           {Object.keys(dats).map(key => (
-            <TableRow
-              dat={dats[key]}
-              key={key}
-              shareDat={shareDat}
-              onDeleteDat={onDeleteDat}
-              onTogglePause={onTogglePause}
-              inspectDat={inspectDat}
-            />
+            <TableRowContainer key={key} datKey={key} />
```
martinheidegger: `key` and `datKey` are the same, seems like one not necessary?

aks-: well the first `key` property is used for the react to use under the hood and is not exposed out to be used in TableRowContainer itself. So we have to send in the second one which we can use to get the dat info in the container.

martinheidegger: I didn't know about [keys](https://reactjs.org/docs/lists-and-keys.html) in React. This opens an entirely different bottle of issues. Come to think of it I think this means we should reserve the `key` variable name to its use in react lists. (and rename other instances where we called something `key` to avoid confusion. I think It would be a good start to then rename the `key` variable [here](https://github.com/dat-land/dat-desktop/pull/506/files#diff-202aca0331bdf210146b5b04c8b74322R18) to `datKey`?

aks-: I think it's okay to use `key` as parameter to functions and NOT as `prop` to any react component. Worth a discussion though. Also if we choose to replace key with datKey, it should be in another issue because we are using it in many different places other than this PR

@juliangruber thoughts?^

juliangruber: I don't see an issue with `key` in JSX meaning a different thing than `key` in JS. And this is the reason why before I would pass around `dat={dat}` to React elements, in order not to have to introduce a `datKey` variable

juliangruber:
very nice!

PR #507, opened 2018-02-07
Improvements on persist state

martinheidegger:
While reviewing a few things on #500 I found a few things and i found it to be shorter to put them in pull requests than to try explain them in comments, the changes I've made:

1. Using locks for writing on disk. This way two events triggered don't run into the issue of taking assumptions of a file while writing to it.
2. Using async/await when writing to disks. Eating the error is really uncomfortable.
3. Turing async method to promises (Dat)
4. A small performance update on writing to disk.

Not sure all of this is useful but maybe it helps.

### DIFF
--- a/app/actions/index.js
+++ b/app/actions/index.js
@@ -11,6 +11,7 @@ import { basename } from 'path'
 import { createLock, createLocker } from 'flexLock'
 
 const dats = {}
+let pausedDats = {}
 const indexLock = createLock()
 const datsLock = createLocker()
 
@@ -176,11 +177,13 @@ const updateConnections = dat => dispatch => {
 
 const updateState = dat => {
   const key = encode(dat.key)
-  const state = !dat.network
+  const state = pausedDats[dat.key]
     ? 'paused'
-    : dat.writable || dat.progress === 1
-      ? 'complete'
-      : dat.network.connected ? 'loading' : 'stale'
+    : !dat.network
+      ? 'disconnected'
+      : dat.writable || dat.progress === 1
+        ? 'complete'
+        : dat.network.connected ? 'loading' : 'stale'
   return { type: 'DAT_STATE', key, state }
 }
 
@@ -207,6 +210,11 @@ export const cancelDeleteDat = () => ({ type: 'DIALOGS_DELETE_CLOSE' })
 export const togglePause = ({ key, paused }) => async dispatch => {
   const unlock = await datsLock(key)
   const { dat } = dats[key]
+  if (paused) {
+    pausedDats[key] = true
+  } else {
+    delete pausedDats[key]
+  }
   if (paused) {
     joinNetwork(dat)(dispatch)
   } else {
@@ -249,14 +257,14 @@ export const loadFromDisk = () => async dispatch => {
   try {
     blob = await readFile(`${homedir()}/.dat-desktop/paused.json`, 'utf8')
   } catch (_) {}
-  const paused = JSON.parse(blob)
+  pausedDats = JSON.parse(blob)
 
   for (const key of Object.keys(datOpts)) {
     const opts = JSON.parse(datOpts[key])
     addDat({
       key: key,
       path: opts.dir,
-      paused: paused[key],
+      paused: pausedDats[key],
       ...opts
     })(dispatch)
   }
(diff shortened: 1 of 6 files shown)


aks-:
@martinheidegger makes sense. Also you likely forgot to run the npm test at the end for standard errors

martinheidegger:
There is an error in the travis push run but I don't understand why. Seems like a bug in travis.

martinheidegger:
I added another major commit to this. Before #9c7138f it stored the information if it was stored as result of it being connected. But actually it is not related. You can loose the connection for a variety of reasons and most of those don't matter to the paused state.

app/actions/index.js
```diff
@@ -1,16 +1,21 @@
...
 import { homedir } from 'os'
 import { clipboard, remote } from 'electron'
 import mirror from 'mirror-folder'
 import fs from 'fs'
 import promisify from 'util-promisify'
 import { basename } from 'path'
+import { createLock, createLocker } from 'flexLock'
 
 const dats = {}
+let pausedDats = {}
```
juliangruber: why create an extra object for this? we don't need to keep track of this state

app/actions/index.js
```diff
@@ -1,16 +1,21 @@
...
+const indexLock = createLock()
+const datsLock = createLocker()
 
+const Dat = promisify(DatRaw)
```
juliangruber: Nice!

app/actions/index.js
```diff
@@ -32,11 +37,11 @@ export const createDat = () => dispatch => {
   addDat({ path })(dispatch)
 }
 
-export const addDat = ({ key, path, paused, ...opts }) => dispatch => {
+export const addDat = ({ key, path, ...opts }) => async dispatch => {
```
juliangruber: we need to be able to add immediately paused dats, because when you pause a dat and restart the app it's still paused

app/actions/index.js
```diff
@@ -32,11 +37,11 @@ export const createDat = () => dispatch => {
...
   if (key) key = encode(key)
   if (!path) path = `${homedir()}/Downloads/${key}`
+  const unlock = await datsLock(key)
```
juliangruber: did you run into any locking issues here before? Instead we could tweak this function to return early if a dat with given key already exists, this would be our "lock"

app/actions/index.js
```diff
@@ -166,16 +176,19 @@ const updateConnections = dat => dispatch => {
...
 const updateState = dat => {
   const key = encode(dat.key)
-  const state = !dat.network
+  const state = pausedDats[dat.key]
     ? 'paused'
-    : dat.writable || dat.progress === 1
-      ? 'complete'
-      : dat.network.connected ? 'loading' : 'stale'
+    : !dat.network
+      ? 'disconnected'
```
juliangruber: if a dat doesn't have a network, it's paused. Why add an extra duplicate state?

app/actions/index.js
martinheidegger: It is easier to keep it in storage and it kinda disconnects the pause-state from the dats (referring to the discussion in the chat)

app/actions/index.js
martinheidegger: `loadFromDisk` loads `pausedDats` before adding all dats dats, this way if they are paused at startup.

app/actions/index.js
martinheidegger: I wrapped it as a safety measure but didn't run into issues yet. Didn't think of returning a dat immediately 😊That might be possible (and better) but there is also the `storeOnDisk` operation way down the road that should be done as well. Which maybe, makes a lock reasonable anyways (think on it?)

app/actions/index.js
martinheidegger: https://github.com/dat-land/dat-desktop/pull/507#issuecomment-363664012

app/actions/index.js
juliangruber: cam we talk about locks in a separate issue? this is trying to do quite a lot and makes it harder to discuss

app/actions/index.js
juliangruber: Why would you close the connection? In dat node, `dat.joinNetwork()` will create `dat.network` immediately and it will only be removed if you call `dat.leaveNetwork()`

martinheidegger: If joinNetwork fails it was still open, had an issue with this.

app/actions/index.js
martinheidegger: I am added the lock on data because I added locking but I think the lock on the index should be part of the persistence PR

martinheidegger: I am added the lock on data because I added locking but I think the lock on the index should be part of the persistence PR

martinheidegger: I am added the lock on data because I added locking but I think the lock on the index should be part of the persistence PR

martinheidegger: I am added the lock on data because I added locking but I think the lock on the index should be part of the persistence PR

martinheidegger: I am added the lock on data because I added locking but I think the lock on the index should be part of the persistence PR

app/actions/index.js
juliangruber: Interesting, I haven't experienced that yet. Shouldn't dat-node throw in this case?

martinheidegger:
Closing this PR in order to prepare a smaller PR.

PR #509, opened 2018-02-07
make title field editable

aks-:


app/components/title-field.js
```diff
@@ -0,0 +1,148 @@
...
+    if (isEditing && writable) {
+      return (
+        <div>
+          <Overlay onClick={() => this.deactivateTitleEditing()} />
+          <EditableFieldWrapper className='bg-white nt1 nb1 nl1 shadow-1 flex justify-between'>
+            <InputField
+              className='bn f6 pl1 normal w-100'
+              defaultValue={editValue || title}
+              onKeyUp={ev => this.handleKeyup(ev)}
+              innerRef={input => {
```
aks-: why innerRef? check here - https://github.com/styled-components/styled-components/issues/102

martinheidegger: If its hard enough to make a note in a github comment, then its maybe hard/rare enough to add a comment to the code?

app/containers/title-field.js
```diff
@@ -0,0 +1,35 @@
...
+
+  return mapStateToProps
+}
+
+const mapDispatchToProps = dispatch => ({
+  updateTitle: (key, path, editValue) =>
+    dispatch(updateTitle(key, path, editValue)),
+  activateTitleEditing: (title, editable) =>
+    dispatch(activateTitleEditing(title)),
+  editTitle: title => dispatch(editTitle(title)),
```
martinheidegger: I was confused by the name `editTitle`. Maybe `directTitleInput` or so? Tbh. `editTitle` seems like the only action that doesn't make sense while the others seem to fit perfectly.

app/actions/index.js
```diff
@@ -200,3 +202,28 @@ export const dropFolder = folder => async dispatch => {
...
+export const activateTitleEditing = title => ({
+  type: 'ACTIVATE_TITLE_EDITING',
+  title
+})
+
+export const editTitle = title => ({ type: 'EDIT_TITLE', title })
+
+export const updateTitle = (key, path, editValue) => {
+  const filePath = `${path}/dat.json`
+  readFile(filePath)
```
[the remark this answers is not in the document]
aks-: @martinheidegger it requires `archive` for it to work https://github.com/joehand/dat-json/blob/master/index.js#L14 which we are not saving in our dat(s). Additionally in the prior version they did almost same as what we are doing right now in this pr - https://github.com/dat-land/dat-desktop/blob/master/lib/dat-json.js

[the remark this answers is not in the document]
martinheidegger: Not sure :+1: or not but can be fixed in a future pr.

app/containers/title-field.js
aks-: changed the name to `updateTemporaryTitleValue`

app/components/title-field.js
```diff
@@ -0,0 +1,154 @@
...
+            { /* why innerRef in following component? check here - styled-components/styled-components#102 */}
+            <InputField
+              className='bn f6 pl1 normal w-100'
+              defaultValue={editValue || title}
+              onKeyUp={ev => this.handleKeyup(ev)}
+              innerRef={input => {
+                this.titleInput = input
+              }}
+            />
+            {editValue === title ? (
```
aks-: this is only place where we need the `editValue` and that's why we are keeping it in the state using the updateTemporaryTitleValue, if we can get rid of it somehow we can have just one flag in state `isTitleUnderEdit` and get rid of updateTemporaryTitleValue action and UPDATE_TEMPORARY_TITLE_VALUE event as well. Maybe when we refactor out this inline title input field to a package...

app/actions/index.js
```diff
@@ -224,3 +226,31 @@ export const dropFolder = folder => async dispatch => {
...
+})
+
+export const updateTemporaryTitleValue = title => ({
+  type: 'UPDATE_TEMPORARY_TITLE_VALUE',
+  title
+})
+
+export const updateTitle = (key, path, editValue) => {
+  const filePath = `${path}/dat.json`
+  readFile(filePath)
```
juliangruber: should use `async/await`

juliangruber: also this fails if there's no `dat.json` yet

app/components/table-row.js
```diff
@@ -128,18 +129,16 @@ const DeleteButton = ({ ...props }) => (
...
-      {dat.metadata.title || `#${dat.key}`}
-    </h2>
-  </div>
-)
-
 const Row = ({ dat, shareDat, onDeleteDat, inspectDat, onTogglePause }) => (
   <Tr
     onClick={ev => {
-      if (ev.target.tagName === 'SVG' || ev.target.tagName === 'use') return
+      if (
```
juliangruber: can be refactored using an array: `if (['SVG', 'use', 'INPUT'].includes(ev.target.tagName))`

app/containers/title-field.js
```diff
@@ -0,0 +1,36 @@
...
+  const { key } = initialProps.dat
+  const mapStateToProps = state => ({
+    dat: state.dats[key],
+    titleUnderEdit: state.titleUnderEdit
+  })
+
+  return mapStateToProps
+}
+
+const mapDispatchToProps = dispatch => ({
```
juliangruber: this is hard to read, maybe let's add the `{ return ... }` around the function block?

app/components/title-field.js
```diff
@@ -0,0 +1,158 @@
+import React, { Component } from 'react'
+import styled from 'styled-components'
+import Icon from './icon'
+import { Plain as PlainButton, Green as GreenButton } from './button'
```
juliangruber: this can be written shorter using

```js
import { * as Button }  from './button'
```

and then

```js
<Button.Green />
```

martinheidegger: I'd prefer not to have any `*` in the import statements as that makes it unclear which of the parts of `./button` are required in the file by just looking at the header.

app/actions/index.js
martinheidegger: A. Currently it fails silently (no catch handler) and it returns the function before the writing is finished. Not sure if writing this async would look so cool:

```javascript
(async function () {
   const metadata = JSON.parse(await readFile(filePath)
   await writeFile(filePath, JSON.stringify({... metadata, title: editValue})
})()
```
Not soo much better, but I agree: slightly.

B. I didn't mention the fail of missing `dat.json` because in practice the duration of not-having a  `dat.json` is very short. (milliseconds?) - quicker than any user can edit a title. Also I am not sure how the whole functionality should work: If there is no `dat.json` - can you edit the title? What if the dat.json was deleted while editing or we had an write error? Should there be an alert? Shouldn't this be part of the "lock" discussion (-> future work)?

app/components/title-field.js
juliangruber: I see that point, but for buttons this is literally just different colors, so you don't get relevant information out of this

martinheidegger: ?! Just reading it like this: I get the information what buttons are used, because only those two buttons are referenced. If I don't use one of the two buttons the linter will complain about an unused reference.

app/actions/index.js
aks-: I thought about it failing in the way if someone deletes the file in middle of editing, what will happen. But was struggling with design on how and where to show the alert, maybe alert box like when we delete the dat. Should we create an issue or add it to todo list?

app/components/title-field.js
aks-: Hm, not sure if I should go ahead with this one?

juliangruber: ok, then 2>1 => PlainButton it is :)

app/reducers/index.js
```diff
@@ -15,7 +15,11 @@ const defaultState = {
     up: 0,
     down: 0
   },
-  inspect: { key: null }
+  inspect: { key: null },
+  titleUnderEdit: {
```
juliangruber: this name isn't clear to me, `under` is mostly used for location.

Can we call this for example `titleEditInPlace`?

aks-: @juliangruber yeah `titleEditInPlace` sounds better. Making that change now....

martinheidegger: My - horrible - default approach would be `currentlyEditedTitle`. To me both `titleEditInPlace` and `titleUnderEdit` are almost equal so I leave it up to you. Other ideas: `focussedTitle` or `titleEditorField`

juliangruber: `currentlyEditedTitle` works for me too, it's natural +1 but different from all the other state names -1

martinheidegger: Since `titleEditInPlace` is already in this PR. lets go with that 😆

PR #510, opened 2018-02-08
Remove #datbots IRC notifications

joehand:
You've been so busy! It makes #datbots pretty noisy =)


juliangruber:
@martinheidegger i think react is the correct base branch, we haven't done anything on #master yet
@joehand thank you for reporting and i take it as a compliment for sure!

PR #516, opened 2018-02-13
add/download screen

aks-:


app/actions/index.js
```diff
@@ -33,6 +33,119 @@ export const createDat = () => dispatch => {
...
+})
+export const hideDownloadScreen = () => ({ type: 'HIDE_DOWNLOAD_SCREEN' })
+export const cancelDownloadDat = key => dispatch => {
+  key = encode(key)
+  const { dat } = dats[key]
+
+  for (const con of dat.network.connections) {
+    con.removeAllListeners()
+  }
+  dat.stats.removeAllListeners()
```
martinheidegger: While it seems necessary that we do this, I kinda thought that `dat.close()` should do that 🤔

app/actions/index.js
```diff
@@ -33,6 +33,119 @@ export const createDat = () => dispatch => {
...
+    else dat.archive.on('content', walk)
+
+    dat.stats.on('update', stats => {
+      if (!stats) stats = dat.stats.get()
+      dispatch({ type: 'DAT_STATS', key, stats: { ...stats } })
+    })
+
+    dispatch(updateState(dat))
+
+    if (!paused) {
```
martinheidegger: Paused is always false!

app/actions/index.js
```diff
@@ -33,6 +33,119 @@ export const createDat = () => dispatch => {
...
+
+      let metadata = {}
+      try {
+        metadata = JSON.parse(blob)
+      } catch (_) {}
+
+      dispatch({ type: 'DAT_METADATA', key, metadata })
+    })
+
+    const walk = () => {
```
martinheidegger: This seems _(unnecessarily)_ copy & pasted from addDat. Please extract into common code.

app/actions/index.js
juliangruber: that's not common in the node ecosystem

juliangruber: that's not common in the node ecosystem

app/actions/index.js
juliangruber: +1

app/actions/index.js
juliangruber: +1

app/components/download.js
```diff
@@ -0,0 +1,135 @@
...
+            <Label>Files:</Label>
+            <FileList dat={dat} />
+          </div>
+        </div>
+        <DetailFooter className='pa3 flex items-center justify-between bg-white'>
+          <p className='truncate'>Download this Dat now?</p>
+          <div className='flex ml2'>
+            <GreenButton
+              onClick={() => {
+                addDat({ key: dat.key, path: dat.path, paused: false })
```
juliangruber: we should default `paused` to `false`

juliangruber:
this still needs some work, I understand it's not an easy case

aks-:
@juliangruber oh, you'd done it, i was just getting to it 🙌

app/actions/index.js
```diff
@@ -134,8 +128,6 @@ export const downloadSparseDat = ({ key }) => dispatch => {
     dispatch(updateState(dat))
     joinNetwork(dat)(dispatch)
     updateConnections(dat)(dispatch)
-
-    dats[key] = { dat, path, opts }
```
aks-: @juliangruber how do you remove this dat and close it when you click cancel download if you don't keep reference to it somewhere?

juliangruber: you're right, I misread the code

juliangruber: will fix!

juliangruber:
ok, sorry about that, unbroke it now

app/actions/index.js
```diff
@@ -33,6 +33,106 @@ export const createDat = () => dispatch => {
   addDat({ path })(dispatch)
 }
 
+export const showDownloadScreen = key => ({
```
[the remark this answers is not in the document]
aks-: @juliangruber you mean merge the logic of `showDownloadScreen` into `downloadSparseDat`? I kept it separate because they do two different things? one opens up the screen and second one initiates the download of dat info.

aks-:
@martinheidegger i checked it out... you mentioned there issues mentioned below
>1. The preview will download data into a folder - so its not actually a preview.
>2. The Download folder is static (it needs to be dynamic for another PR
>3. Dats can never be downloaded twice because "dat" is our unique, why would we download dats more than once? Versions!

1. it does not download data into the folder, it only downloads metadata to show on download screen.
2. i'm not quite sure if i fully understand this one.
3. i agree, but we don't have a way to tell users as of now that they are trying to download an existing dat even in our master. this pr behaves similar to what is there on master. (cc: @Kriesse)

karissa:
@aks- on point 3, I guess you could iterate through existing dats and see if one still exists with that id.

aks-:
@martinheidegger can you please take a look at this one again? i think we can merge it

martinheidegger:
@aks- My review is contained in #529 and #530 respectively.

martinheidegger:
Now this PR contains problems if the same dat is downloaded twice. I think that is the last issue before we should merge it to react?!

soyuka:
@martinheidegger should I review this? Let me know!

app/actions/dat-middleware.js
```diff
@@ -0,0 +1,423 @@
...
+    // Telling it to all the middlewares.
+    this.dispatch(action)
+    exec(action)
+      .then(data =>
+        this.dispatch({ ...action, type: `${action.type}_SUCCESS`, ...data })
+      )
+      .catch(error =>
+        this.dispatch({ ...action, type: `${action.type}_ERROR`, error })
+      )
+    return true
```
soyuka: We can use `async/await` here.

tests/index.js
```diff
@@ -1,120 +1,162 @@
+#!/usr/bin/env node
```
soyuka: can't we use async/await in those tests? IMO it'd make things more readable.

app/actions/dat-middleware.js
```diff
@@ -0,0 +1,423 @@
...
+      REMOVE_DAT: action => this.removeDat(action),
+      TRY_ADD_DAT: action => this.tryAddDat(action),
+      REQUEST_DOWNLOAD: action => this.validateDownloadRequest(action),
+      TOGGLE_PAUSE: action => this.togglePause(action),
+      DOWNLOAD_SPARSE_DAT: action => this.downloadSparseDat(action),
+      CANCEL_DOWNLOAD_DAT: action => this.cancelDownloadDat(action)
+    }
+  }
+
+  async validateDownloadRequest ({ key }) {
```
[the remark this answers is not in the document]
martinheidegger: All actions are async, this way it is more consistent (also in error case)

app/actions/dat-middleware.js
martinheidegger: The `return true` statement returns before the `.then().catch()` is finished. As such the following [`dispatch`](https://github.com/dat-land/dat-desktop/pull/516/files#diff-66dae0682d14f11364314215e58de09dR67) is called in parallel to the execution.

tests/index.js
martinheidegger: It is possible. The tests right now are nothing but a move of the tests of master to this branch.

martinheidegger:
@soyuka there are still several, severe known bugs here 😅.

martinheidegger:
So: I gave this another shot. @soyuka It would be nice if you could review this once more. I would love to do a rebase after merging #541 :)

martinheidegger:
@soyuka Thank you for reviewing #541, I merged it in this branch. Can you give this branch another look?

martinheidegger:
@AtuyL thank you for the fixes, would you mind to review this PR officially ;)

tests/utils/waitForMatch.js
```diff
@@ -0,0 +1,40 @@
+var wait = require('./wait')
+
+module.exports = function waitForMatch (t, app, selector, regexp, ms, reverse) {
+  if (reverse) {
+    reverse = true
+  } else {
+    reverse = false
+  }
```
soyuka: can be simplified: `reverse = reverse ? true : false`

tests/utils/waitForMatch.js
```diff
@@ -0,0 +1,40 @@
+var wait = require('./wait')
+
+module.exports = function waitForMatch (t, app, selector, regexp, ms, reverse) {
```
soyuka: `ms = 10000` in the function params instead of l9

app/components/button.js
```diff
@@ -3,7 +3,17 @@
...
+const BaseButton = styled.button.attrs({
+  onClick: props => {
+    if (props.trigger) {
+      return function (e) {
+        e.stopPropagation()
+        props.trigger.apply(this, [e])
+      }
+    }
+    return undefined
+  }
```
soyuka: I'd suggest a minor change (reverse condition + use call to avoid declaring an array):

```
props => {
  if (!props.trigger) return
  return function (e) {
      e.stopPropagation()
      props.trigger.call(this, e)
  }
}
```

package.json
```diff
@@ -7,19 +7,19 @@
   "description": "Dat Desktop App",
   "author": "Dat Team <community@datproject.org>",
   "scripts": {
-    "start": "cross-env NODE_V=\"$(node -v)\" NODE_ENV=development electron .",
-    "pretest": "npm run lint",
-    "test": "npm run test:unit",
+    "start": "npm run build:dev && cross-env NODE_ENV=development npm-run-all -p start:watch start:electron",
```
AtuyL: This script-command(prestart) is not recommended ?

`"prestart": "npm run build:dev",`
`"start": "cross-env NODE_ENV=development npm-run-all -p start:watch start:electron",`

martinheidegger: `prestart` should work: Do you want to make a new PR?

PR #515, opened 2018-02-13
[WIP] Viewers

juliangruber:
This is a WIP pull request. Looking for early stage feedback.

The basic text viewer works!

PR #519, opened 2018-02-20
React Viewer: Streaming API (and other tweaks)

martinheidegger:
Looking over #515 I wondered how a different API could look like that more future proof to our case. I came up with the API in this PR. Essentially it adds following things:

1. It adds a very simple streaming implementation - in case one just wants to stream the whole file
2. It adds a `initViewer` command to init one viewer; subsequently `viewer:open` events are called which each should be triggered when someone opens the webview - this way the viewer can theoretically be reused in future iterations instead of respawned for each file
3. It uses dat links with versions as file reference which should make sure that the preview always loads the correct data from the correct dat archive. Aside from this helping debugging, it should help to figure out timing problems in the user interface once we implement versions and rapid navigation in the files.
4. It adds a few improvements to error handling (now we get file number and log type for events in the viewer

### DIFF
--- a/app/components/viewer/index.js
+++ b/app/components/viewer/index.js
@@ -1,33 +1,82 @@
 import React from 'react'
 import Dialog from '../dialog'
+import url from 'url'
+
+const streams = {}
 
 const Viewer = ({ show, dat, path, onExit }) => {
+  const datPath = `dat://${dat && dat.key.toString('hex')}/${path}#${dat && dat.version}`
+  const initFile = (datPath, cb) => {
+    const parts = url.parse(datPath)
+    // TODO: version is not considered ...
+    if (parts.hostname !== dat.key.toString('hex')) return cb(new Error('Dat already closed'))
+    dat.archive.stat(parts.path, cb)
+  }
+  const readFile = (datPath, start, end, cb) => {
+    const parts = url.parse(datPath)
+    // TODO: version is not considered ...
+    if (parts.hostname !== dat.key.toString('hex')) return cb(new Error('Dat already closed'))
+
+    // TODO: multiple reads on the same file should not open/close the file
+    dat.archive.open(path, 'r', (err, fd) => {
+      if (err) return cb(err)
+
+      const buf = Buffer.alloc(end - start)
+      dat.archive.read(fd, buf, 0, buf.length, start, err => err ? cb(err) : cb(null, buf))
+    })
+  }
   const onref = webview => {
     webview.addEventListener('ipc-message', ev => {
-      if (ev.channel !== 'req') return
-      const { cmd, args, id } = ev.args[0]
+      if (ev.channel === 'viewer:init') {
+        webview.send('viewer:open', datPath)
+        return
+      }
+      if (ev.channel === 'viewer:stream') {
+        const {cmd, id, datPath, start, end} = ev.args[0]
+        if (cmd === 'open') {
+          const parts = url.parse(datPath)
+          // TODO: version is not considered ...
+          if (parts.hostname !== dat.key.toString('hex')) return webview.send('viewer:stream', id, new Error('Dat already closed'))
 
-      if (cmd === 'read') {
-        const [start, end] = args
-        dat.archive.open(path, 'r', (err, fd) => {
-          if (err) return webview.send('res', id, err.message)
-
-          const buf = Buffer.alloc(end - start)
-          dat.archive.read(fd, buf, 0, end - start, start, err => {
-            if (err) return webview.send('res', id, err.message)
-
-            webview.send('res', id, null, buf.toString('hex'))
+          const stream = dat.archive.createReadStream(parts.path, {start, end})
+          stream.on('data', data => webview.send('viewer:stream', id, null, data.toString()))
+          stream.on('error', error => webview.send('viewer:stream', id, error))
+          stream.on('end', data => {
+            delete streams[id]
+            webview.send('viewer:stream', id, null, null)
           })
-        })
-      } else if (cmd === 'size') {
-        dat.archive.stat(path, (err, stat) => {
-          if (err) return webview.send('res', id, err.message)
-          webview.send('res', id, null, stat.size)
-        })
+          streams[id] = stream
+        }
+        if (cmd === 'close') {
+          const stream = streams[id]
+          if (stream) stream.close()
+        }
+        return
+      }
+      if (ev.channel === 'viewer:exec') {
+        const {cmd, id, args} = ev.args[0]
+        const cb = (err, data) => {
+          webview.send('viewer:exec', id, err, data)
+        }
+        if (cmd === 'init') {
+          const [datPath] = args
+          initFile(datPath, cb)
+        } else if (cmd === 'read') {
+          const [datPath, start, end] = args
+          readFile(datPath, start, end, cb)
+        }
       }
     })
     webview.addEventListener('console-message', e => {
-      console.log(e.message)
+      var srcFile = e.sourceId.replace(/^.*[\\/]/, '')
+      const msg = `[webview@${srcFile}(${e.line})]: ${e.message}`
+      if (e.level === 2) {
+        console.error(msg)
+      } else if (e.level === 1) {
+        console.warn(msg)
+      } else {
+        console.log(msg)
+      }
     })
   }
   return (
--- a/app/components/viewer/preload.js
+++ b/app/components/viewer/preload.js
@@ -1,21 +1,261 @@
 const { ipcRenderer } = require('electron')
 
-const callbacks = []
+let currentId = 0.0
 
-window.read = (start, end, cb) =>
-  send('read', [start, end], (err, str) => {
-    if (err) return cb(err)
-    cb(null, Buffer.from(str, 'hex'))
-  })
-window.size = cb => send('size', [], cb)
+/*
+;['log', 'warn', 'error', 'info'].forEach(function (prop) {
+  const old = console[prop]
+  console[prop] = function () {
+    let arg = arguments[0]
+    if (arguments.length > 1) {
+      arg = Array.from(arguments).join(' ')
+    }
+    old.call(console, arg)
+  }
+})
+*/
 
-const send = (cmd, args, cb) => {
-  const id = callbacks.length
-  callbacks.push({ id, cb })
-  ipcRenderer.sendToHost('req', { cmd, args, id })
+function nextId () {
+  currentId += 1.0
+  return currentId
 }
 
-ipcRenderer.on('res', (_, id, err, ...args) => {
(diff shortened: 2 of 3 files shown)


juliangruber:
extracting the api into this comment:

```js
addEventListener('viewer:open', aysnc ({ file }) => {
  openFile(file)
  file.createReadStream() // ...
  await file.read(start, end) // ...
  file.destroy()
})
initViewer()
```

Questions:

- why do you need to call `openFile(file)`?
- why do you need to call `initViewer()`?
- what about a node style api like this:

```js
let current // ...

module.exports =  aysnc ({ file }) => {
  current = file
  file.createReadStream() // ...
  await file.read(start, end) // ...
  file.destroy()
})
```

Comments:

- great idea to add `.createReadStream()`!

martinheidegger:
I chose this approach with following thoughts in the background:

1. I didn't know if it was possible to implement a plugin API that uses CommonJS modules `module.exports`. But even if it was: a lot of minor UI libraries might take a while to load before they can be used  _(video player, etc.)_ as it might be necessary to deal with html/css loading as well. For that reason I found a dedicated, required `initViewer` the safest bet. Indeed, the Viewer will not work without calling `initViewer` but any choice of library will work. _(Note: I am not entirely happy with the naming, I thought about `ready()` or `activate`)_
2. I split up `initViewer` and the file-listening because I wanted to make it possible to pre-load each viewer without a file present.
3. An additional problem I always see with an `file => {}` API is that the implementor is not necessarily clear on the memory ownership of `file`. The reference might be used outside of the viewer implementation. To me it always **feels** like I am not responsible to clean it up.... But with `file = openFile()` it is usually clear that I am the only person using that instance. I nearly named it `createFile` or just `new File` to make it even clearer but I thought that `openFile` is more appropriate.

app/viewers/text/index.js
```diff
@@ -2,10 +2,31 @@ const pre = document.createElement('pre')
...
-    if (err) throw err
-    pre.innerText = buf.toString()
+let current
+let firstData
+
+window.addEventListener('viewer:open', async (event) => {
+  if (current) {
+    current.destroy()
+  }
+  const file = event.file
```
martinheidegger: Maybe this should be `datPath` instead of `file`.

app/components/viewer/preload.js
```diff
@@ -1,21 +1,261 @@
...
+  const old = console[prop]
+  console[prop] = function () {
+    let arg = arguments[0]
+    if (arguments.length > 1) {
+      arg = Array.from(arguments).join(' ')
+    }
+    old.call(console, arg)
+  }
+})
+*/
```
martinheidegger: Logging of errors - right now - doesn't provide a stack trace but at least keeps the last line-number for some debugging. However, the current logging does not take care of the additional arguments, so the block above concats the arguments. We loose the stack of the error though :(

juliangruber:
> I didn't know if it was possible to implement a plugin API that uses CommonJS modules module.exports. But even if it was: a lot of minor UI libraries might take a while to load before they can be used (video player, etc.) as it might be necessary to deal with html/css loading as well. For that reason I found a dedicated, required initViewer the safest bet. Indeed, the Viewer will not work without calling initViewer but any choice of library will work. (Note: I am not entirely happy with the naming, I thought about ready() or activate)

what's the difference between

```js
addEventListener('file', async file => {
  // thing is now ready
})
loadThing(() => initViewer())
```

and

```js
addEventListener('file', async file => {
  if (!thing.ready) await loadThing()
})
```

This seems to be an implementation detail and shouldn't be favoured over a simpler API design.

juliangruber:
> I split up initViewer and the file-listening because I wanted to make it possible to pre-load each viewer without a file present.

I don't understand this yet, in which scenario is the file not yet present? In my understanding you click on a file in the list, then the viewer opens and we pass it the file instance. Do you instead want to initialise all the viewers when starting the app?

juliangruber:
> An additional problem I always see with an file => {} API is that the implementor is not necessarily clear on the memory ownership of file. The reference might be used outside of the viewer implementation. To me it always feels like I am not responsible to clean it up.... But with file = openFile() it is usually clear that I am the only person using that instance. I nearly named it createFile or just new File to make it even clearer but I thought that openFile is more appropriate.

The plugin runs in a `<webview>` so can be disposed at any time

martinheidegger:
> what's the difference between
>
> ```javascript
> // Ⓐ
> addEventListener('file', async file => {
>   // thing is now ready
> })
> loadThing(() => initViewer())
> ```
>
> 
> ```javascript
> // Ⓑ
> addEventListener('file', async file => {
>  if (!thing.ready) await loadThing()
>  // thing is now ready
> })
> ```

This is related to
> @martinheidegger: ... make it possible to pre-load each viewer

and

> @juliangruber: I don't understand this yet, in which scenario is the file not yet present?

Loading a viewer might take some time: If we do this after the user has clicked an Item it might feel sluggish. We can preload all the viewers at the start of the Dat app or we can do it more intelligently at the time of looking at a file-tree _(and preload only the viewers that might be clicked)_. In each case It is not-verbose that the "Ok, we are ready to go"-Signal  (`initViewer`) is triggered by the "I will do this with every file"-implementation (`addEventListener`). I **did** think of putting both together but the initiation is something done once, while the reading of files _(its called **listener**)_ is supposed to be called multiple times. It is a less stringent API but imo. it communicates the purpose better.

> The plugin runs in a <webview> so can be disposed at any time, as well as being separated from the process space of the main window.

Of course: an outside deletion/closing can always occur. _(afaik. there are browser events that already let the implementation deal with that case)_. The problem is more about long-living viewers. Those viewers need to destroy the instances they receive else they run into memory leaks. With `file => {}` they might assume that each time they receive a new file event: the former one gets taken care of by the outside process. By making it verbose it is my hope that people don't forget about it 😅 .

juliangruber:
Is this premature optimization? As long as we can add a preloading mechanism without breaking the API, I don't think this is relevant yet. We could later on add optionally:

```js
exports.preload = async () => {
  // preload everything here
}
```

app/components/viewer/index.js
```diff
@@ -1,33 +1,82 @@
...
-          if (err) return webview.send('res', id, err.message)
+    // TODO: multiple reads on the same file should not open/close the file
+    dat.archive.open(path, 'r', (err, fd) => {
+      if (err) return cb(err)
 
-          const buf = Buffer.alloc(end - start)
-          dat.archive.read(fd, buf, 0, end - start, start, err => {
-            if (err) return webview.send('res', id, err.message)
+      const buf = Buffer.alloc(end - start)
+      dat.archive.read(fd, buf, 0, buf.length, start, err => err ? cb(err) : cb(null, buf))
```
aks-: @martinheidegger uh looks like you forgot to run `npm test`

aks-:
@juliangruber 
>As long as we can add a preloading mechanism without breaking the API
so what you're saying is preload all the viewer plugin(s) in `exports.preload` and get rid of initViewer?

app/components/viewer/preload.js
```diff
@@ -1,21 +1,261 @@
...
+}
+
+class Stream extends EventEmitter {
+  constructor (file, start, end) {
+    super()
+    this.id = nextId()
+    this.file = file
+    this.start = start
+    this.end = end
+    this.removeAllListeners()
```
aks-: @martinheidegger why this exists?

app/components/viewer/preload.js
```diff
@@ -1,21 +1,261 @@
...
+    validateRange(file, start, end)
+      .then(({start, end}) =>
+        ipcRenderer.sendToHost('viewer:stream', {
```
aks-: Why is there IPC in a stream mechanism's foundation? I mean it's fine but looks like mixing up two things...

app/components/viewer/preload.js
juliangruber: error {de,}serialization is often a pain, this might be helped by using a module like https://www.npmjs.com/package/serialize-error

app/components/viewer/preload.js
juliangruber: true, I'm curious about this too

app/components/viewer/preload.js
juliangruber: I think this tells the host that this stream is now ready to receive and forward data

juliangruber:
> @juliangruber
> 
> As long as we can add a preloading mechanism without breaking the API
> 
> so what you're saying is preload all the viewer plugin(s) in exports.preload and get rid of initViewer?

Hm hm hm we need to think about preloading a little more. It feels risky to preload foreign code on app start, even if it runs in a separate context.

@martinheidegger for the Viewers you have in mind, how long will they take to preload?

app/components/viewer/preload.js
martinheidegger: In the current EventEmitter implementation `.removeAllListeners()` (re-)initializes the object and specifies what events are allowed:

https://github.com/dat-land/dat-desktop/blob/9dc26c2d3ca82adff8c4cde8b65675f91971484c/app/components/viewer/preload.js#L136-L142

and 

https://github.com/dat-land/dat-desktop/blob/9dc26c2d3ca82adff8c4cde8b65675f91971484c/app/components/viewer/preload.js#L80-L82

app/components/viewer/preload.js
martinheidegger: Looks interesting, ... https://www.npmjs.com/package/serialize-error

martinheidegger:
> It feels risky to preload foreign code on app start

It can be anytime when there have free resources, doesn't need to be directly at app-start. I am not sure how long it takes, but it could easily be between 1~5 seconds; Not much but it sums up for various viewers.

app/components/viewer/preload.js
martinheidegger: It does?!

app/components/viewer/preload.js
aks-: Is it worth separating this so this stream class is usable in other cases?

app/components/viewer/preload.js
juliangruber: I meant it should happen in `EventEmitter`'s constructor, not really a stream's responsibilities

martinheidegger: Hmm, I could pass a set of allowed events to the EventEmitter, would you be okay with that?

martinheidegger:
@juliangruber do you feel comfortable if we continue this discussion after merging this part in #515

soyuka:
> Naturally, since I opened this PR, I am all for this. So please expand on it!

:+1: let's see how @juliangruber feels about that!

> I would like to release the react branch before working on this.

I'm definitely :+1: on this, I want to try out the react branch more thoroughly and will review it later this week! The viewer feature is a good improvement but it's not a requirement to get the react branch released IMO.

PR #520, opened 2018-02-27
Add/viewers streaming

aks-:


app/components/viewer/preload.js
```diff
@@ -1,4 +1,5 @@
 const { ipcRenderer } = require('electron')
+const { EventEmitter2 } = require('eventemitter2')
```
juliangruber: Why not use `require('events').EventEmitter`?

aks-: I didn't realize that was already coming in as a dependency, used that

PR #522, opened 2018-03-05
[WIP] - add non-dom based unit tests

aks-:


package.json
```diff
@@ -8,18 +8,24 @@
   "author": "Dat Team",
   "scripts": {
     "start": "electron .",
-    "test": "prettier-standard 'app/**/*.js' 'tests/**/*.js' 'lib/**/*.js' && standard",
-    "watch": "webpack --watch"
+    "test": "prettier-standard 'app/**/*.js' 'tests/**/*.js' 'lib/**/*.js' 'unit-tests/*.js' && npm run unit-tests && standard",
```
martinheidegger: Can we split lint & unit tests? Also: Order <lint> && <unit> && <lint> is weird.

unit-tests/hex-content.js
```diff
@@ -0,0 +1,67 @@
+import test from 'tape'
+import React from 'react'
+import configure from 'enzyme-adapter-react-helper'
+import { shallow } from 'enzyme'
+import HexContent from '../app/components/hex-content'
+
+configure()
```
martinheidegger: Refactoring this to something like:

```javascript
import React from 'react'
import { init, shallow } from `./lib/enzymed`
import HexContent from '../app/compoents/hex-content'

const test = init() // runs the enzyme config and returns tape
```

would make it harder to make a mistake in the test setup

unit-tests/hex-content.js
```diff
@@ -0,0 +1,67 @@
...
+  const wrapper = shallow(
+    <HexContent
+      dat={{
+        state: 'loading',
+        paused: true
+      }}
+    />
+  )
+
+  t.equal(wrapper.find('.color-blue.hover-color-blue-hover').length, 2)
```
martinheidegger: Just a thought but: `wrapper.find('.hexagon').hasClass('.color-blue')` seems a lot more readable ?!

unit-tests/inspect.js
```diff
@@ -0,0 +1,149 @@
...
+        },
+        path: 'A-path'
+      }}
+      closeInspectDat={() => {}}
+      addFilesToDat={() => {}}
+    />
+  )
+
+  t.equal(
+    wrapper
```
martinheidegger: Cant you extract the `wrapper.find('.is-selectable')` into a variable? i think it would be much easier to read.

unit-tests/hex-content.js
aks-: hm, i was thinking of doing something like `import { shallow, render } from '../helpers/enzyme';` that is a helper file that does `configure()` &c and re-exports enzyme methods. But I thought it was not a big deal hmm

i'll do it in separate commit 👍

unit-tests/inspect.js
aks-: yep good catch. i'll do it 👍

package.json
aks-: yes i agree, i'll split them and fix it.

unit-tests/_helpers/enzymeSetup.js
```diff
@@ -0,0 +1,3 @@
+import configure from 'enzyme-adapter-react-helper'
```
martinheidegger: 👍

aks-:
>It is tricky and a case-by-case judgement what section should be tested and what shouldn't. Maybe it would be good to find a short guideline (added to Contributing.md) that explains how to judge this best?

@martinheidegger test what changes, and no tautological tests? Not sure if that needs to be added to contributing.md... maybe we can document how test env is structured, because it would require us to mock objects like `window` etc which i'm thinking of doing in `unit-tests/_helpers`

@juliangruber thoughts^

PR #523, opened 2018-04-09
Readded info-screen

martinheidegger:
In the react branch the info-screen (hamburger button) of the master branch wasn't added. This PR adds the info-screen with the Report-bug button.


app/reducers/index.js
```diff
@@ -314,6 +318,14 @@ const redatApp = (state = defaultState, action) => {
           }
         }
       }
+    case 'TOGGLE_MENU':
```
aks-: do you think it would make more sense to call it `TOGGLE_BUG_MENU` instead?

martinheidegger: Right now it only shows the debug information but I believe it was added to hold more information.

aks-: aha i see. looks good,  i think we can merge it then

PR #529, opened 2018-04-16
Extracted actions into middleware

martinheidegger:
In a future PR I need to specify the home director for DAT operations. This PR also moves globals of actions (`dats`) from [a global variable](https://github.com/dat-land/dat-desktop/blob/af328a1d5ca2a3dcfcfc3959d3bc42832accf2b0/app/actions/index.js#L12) into a [class/instance system](https://github.com/dat-land/dat-desktop/blob/a5e2aa180e689a34789bbce91ae94af7afa915f9/app/actions/dat-middleware.js#L19). This also turns the [`LOAD_ACTION`](https://github.com/dat-land/dat-desktop) - (_which wasn't a real action_) in [`./app/index.js`](https://github.com/dat-land/dat-desktop/blob/af328a1d5ca2a3dcfcfc3959d3bc42832accf2b0/app/index.js#L26) into an [`regular operation`](https://github.com/dat-land/dat-desktop/blob/a5e2aa180e689a34789bbce91ae94af7afa915f9/app/index.js#L31).

It also theoretically allows multiple react windows interacting with the same middleware through [a very basic listener system](https://github.com/dat-land/dat-desktop/blob/a5e2aa180e689a34789bbce91ae94af7afa915f9/app/actions/dat-middleware.js#L60-L62).

It furthermore adds the [success/error concept](https://github.com/dat-land/dat-desktop/blob/add/download-middleware/app/actions/dat-middleware.js#L36-L42) to all middleware operations. My thinking is that there should be a `{ACTION}`-action before `{ACTION}_SUCCESS` or `{ACTION}_ERROR` if an error/success happened during the operation. This would allow to show errors if errors occur and progress displays in the time between an action starts and an action finishes.

### DIFF
--- a/app/actions/dat-middleware.js
+++ b/app/actions/dat-middleware.js
@@ -1,0 +1,380 @@
+'use strict'
+
+import Dat from 'dat-node'
+import { encode } from 'dat-encoding'
+import { homedir } from 'os'
+import fs from 'fs'
+import { basename } from 'path'
+import { ipcRenderer, shell } from 'electron'
+import mirror from 'mirror-folder'
+import promisify from 'util-promisify'
+
+const readFile = promisify(fs.readFile)
+const writeFile = promisify(fs.writeFile)
+const mkdir = promisify(fs.mkdir)
+const { Notification } = window
+
+export default class DatMiddleware {
+  constructor () {
+    this.dats = {}
+    this.listeners = []
+    this.execByType = {
+      UPDATE_TITLE: action => this.updateTitle(action),
+      REMOVE_DAT: action => this.removeDat(action),
+      TRY_ADD_DAT: action => this.tryAddDat(action),
+      TOGGLE_PAUSE: action => this.togglePause(action),
+      DOWNLOAD_SPARSE_DAT: action => this.downloadSparseDat(action),
+      CANCEL_DOWNLOAD_DAT: action => this.cancelDownloadDat(action)
+    }
+  }
+
+  execAction (action) {
+    const exec = this.execByType[action.type]
+    if (!exec) return false
+    // Telling it to all the middlewares.
+    this.dispatch(action)
+    exec(action)
+      .then(data =>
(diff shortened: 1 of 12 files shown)


aks-:
@martinheidegger i haven't checked the code yet but tried to use the app with this pr, and i've run into few issues..

-  i'm unable to delete the dat.
-  behaviour of pause/resume dat is non-deterministic in the sense that it won't pause the download of dat at very moment and sometimes lag if i click the pause/resume button in relatively small time span.

also do you think it would be good idea for us to refactor actions etc once we have deployed the react on master? we don't have any tests for actions in place for now so i guess we could do that once we deploy this and then start on with the refactoring?

martinheidegger:
As mentioned in gitter. This PR is necessary to re-add CLI-options that exist on master and make sure that the integration tests work again. This PR is just a precursor to follow-up work. That being said: It should be possible to delete dats and pause/play behavior should be deterministic.

The PR was mostly nothing but a move operation of the actions to the middleware. As far as I understood I didn't change any logic. This means that the bugs could be refactoring bugs or existed before. In any case: thanks for testing it, let me investigate!

martinheidegger:
I fixed the download bug in #5da2ff8 and found another issue in #02444df. But I couldn't reproduce your temporal issues with the dats. My suggestion would be to merge this as-is and try to figure out how to make the pause/resume behavior more stable.

martinheidegger:
@aks- Found the issue in pause/resume state, should be addressed in https://github.com/dat-land/dat-desktop/pull/529/commits/eb08fb1b60af665491a549e836260b65519e7398

aks-:
@martinheidegger deleting the dat is failing on this one

aks-:
@martinheidegger should we close this in favour of https://github.com/dat-land/dat-desktop/pull/530?

martinheidegger:
Following our discussion in gitter I merged this. To your question: I tend to build on existing PR with new PR if the first ones don't get merged quickly. The expectation on my side is that the first PR (this) gets merged before the discussion on the second PR (530) begins.

PR #533, opened 2018-05-18
Dependency updates

martinheidegger:
This PR updates the dependencies in the react branch, making sure we are up-to-date.

_Notes:_
- I used bundle.js instead of build.js for the webpack build because its what seems standard assumes as default.
- When updating webpack for some reason the bundle wasn't written without setting a `path` property.
- Updating electron required using `hiddenInset` instead of `hidden-inset`.
- `react-file-drop` now assumes a runtime with `es-modules` available so I had to run it through Babel.
- This PR fixes 596 vulnerability issues. (One still persists)


webpack.config.js
```diff
@@ -2,13 +2,17 @@ const nodeExternals = require('webpack-node-externals')
...
   entry: './app/index.js',
-  target: 'electron',
-  externals: [nodeExternals()],
+  target: 'electron-main',
+  externals: [nodeExternals({
+    whitelist: /react-file-drop/
+  })],
   output: {
-    filename: 'static/build.js',
+    path: `${__dirname}/static`,
```
karissa: this won't work on windows, try using `path.join` instead

martinheidegger:
@karissa Should be fixed now!

webpack.config.js
```diff
@@ -1,22 +1,27 @@
 const nodeExternals = require('webpack-node-externals')
+const path = require('path')
 
 module.exports = {
-  entry: './app/index.js',
-  target: 'electron',
-  externals: [nodeExternals()],
+  entry: path.join('.', 'app', 'index.js'),
```
soyuka: for consistency can we replace `.` by `__dirname`? 

Also, because we know the path we can write `path.normalize('./app/index.js')` so that it works on windows (same below 
```path.normalize(`${__dirname}/static`)```) [`path.join` calls `normalize`](https://github.com/nodejs/node/blob/master/lib/path.js#L490).

soyuka:
Small comment but LGTM for package updates

webpack.config.js
martinheidegger: Oh, just read the code and it also uses normalize on mac. Learning new things every day. (thx)

martinheidegger:
@soyuka updated the PR

soyuka:
Thanks @martinheidegger !

PR #535, opened 2018-06-06
Updates in the development process

martinheidegger:
This PR readds most if all scripts available in the master branch to the react branch and improves on them and a few developer bits and pieces; listing them up:

- Security improvements _(following: [the electron guide](https://electronjs.org/docs/tutorial/security) and [eslint recommendations](https://eslint.org/docs/rules/no-implied-eval))_ − 43acfb8
- Setting the `mode` for build in npm scripts allowing for `development` and 
`production` builds. − 36ed83d
- Open devtools only in developer mode, to make sure that they are not opened in production - 97d98f4
- Added react developer tools and the `update-rdt` script to update the tools − 32b0a93
- Readded dist & pack scripts − 28ad14f
- Added Mac App Icons to package. − 6561e75
- Added Linux build support _(requires DAT team email address!)_ − 052cfe1
- Added Window build support − c55ae72
- Fixed Environment variables in development process _(using [`cross-env`](https://www.npmjs.com/package/cross-env) and [`DefinePlugin`](https://webpack.js.org/plugins/define-plugin/)) to make sure that the environment variables will be respected in the build scripts)_ − a77a712
- Fix: Auto-Updater can thrown an error, catching and returning it. − b62ffd9
- Readded clean & burnthemall scripts, though in a slightly different fashion _(now using `git clean` instead of `rm` to clean *all* ignored files.)_ − ded7be1
- Readded deps script − 1dd3e43
- Updated deep-dependencies using `burnthemall` − 1e849d6

I might have forgotten a detail here and there but this should bring us closer to a react release.


soyuka:
Good improvements, thanks for working on this!

Do we have to publish react-dev-tools as part of the repository? Can't it be installed with npm instead?

martinheidegger:
I tried using npm versions but they didn't work (too old). Without a proper release on NPM, other options seemed ... bothersome.

martinheidegger:
`react-devtools` doesn't have a `manifest.json` which means it can not be added to electron.

package.json
```diff
@@ -5,27 +5,49 @@
...
-    "test:unit": "NODE_ENV=test babel-tape-runner -r ./unit-tests/_helpers/*.js unit-tests/*.js",
+    "test:unit": "cross-env NODE_ENV=test babel-tape-runner -r ./unit-tests/_helpers/*.js unit-tests/*.js",
     "lint": "prettier-standard 'app/**/*.js' 'tests/**/*.js' 'lib/**/*.js' 'unit-tests/*.js' && standard",
-    "build": "webpack",
-    "watch": "webpack --watch"
+    "deps": "cross-env DEBUG=* dependency-check . --detective precinct --entry app/index.js",
+    "clean": "git clean -dfX",
+    "burnthemall": "rm -rf ~/.electron && npm run clean && rm -f package-lock.json && npm install",
+    "update-rdt": "rm -rf dev/react-dev-tools && ced fmkadmapgofadopljbjfkapdkoienihi dev/react-dev-tools",
+    "dev": "cross-env NODE_ENV=development sh -c \"(npm run watch &); npm run start\"",
```
soyuka: won't this lead to zombie processes? I mean how do you stop watching? 

I think that there's a way to launch electron from webpack so that we would only need `npm run start`.

webpack.config.js
```diff
@@ -31,5 +30,10 @@ module.exports = {
         }
       }
     ]
-  }
-}
+  },
+  plugins: [
+    new DefinePlugin({
+      'process.env.NODE_ENV': argv.mode
```
soyuka: `argv.mode || 'development'`? Maybe it already has a default value though!

martinheidegger: https://webpack.js.org/concepts/#mode - default is "production".

package.json
martinheidegger: It indeed creates a zombie process. Thanks for noticing and pointing out. I am looking now into https://webpack.electron.build/ ー It would change quite a few things.

martinheidegger:
I saw your :confused: icon-comment. The only way I found to add the react-dev-tools was through [`BrowserWindow.addDevToolsExtension`](https://github.com/dat-land/dat-desktop/pull/535/files#diff-168726dbe96b3ce427e7fedce31bb0bcR31). And in order to add it the folder needs to contain a full chrome extension (which includes the `manifest.json`). The [content of the `react-devtools` package](https://unpkg.com/react-devtools@3.2.3/) does not contain the entire extension. Skimming through the docs I didn't find another way to add the extension to Electron. There is the [`electron-react-devtools`](https://www.npmjs.com/package/electron-react-devtools) package but it is outdated and doesn't work as advertised.

soyuka:
Thanks for the detailed explanation. It's sad that there are no "official" way of adding react-devtools to something as popular as electron...

martinheidegger:
@soyuka I took some time and looked into `electron-webpack` and I don't think its a good fit. Not now and maybe not in future, because:

- It uses a specific folder layout which is not the same we use, we would need to reorganize our code which would result in breaking PR's
- It is entirely built on babel 7 - still beta
- It doesn't support preload scripts (which this PR uses: https://github.com/dat-land/dat-desktop/commit/43acfb87d988b40aecae545f7d9e5bdc0877eeba)

In an attempt to "just use it" I have struggled to make it work, so instead of trying to fit dat-desktop on `electron-webpack` I added more features to this that should basically fix your issues:

- Now this PR runs start & watch in parallel using `npm-run-parallel` − 1294c5d7ff9
- Added simple watch process on basis of `gaze` that respects `preload` - 9f9252c0d5

and also fixed a few uglies:

- Cleanup: Removed debug console statement - 2530271
- Added documentation on the security aspect of preload.js - 72b6ea3

index.js
```diff
@@ -25,10 +25,12 @@ menu[menu.length - 1].submenu.push({
 })
 
 let win
+let watchProcess
 
 app.on('ready', () => {
   if (isDev) {
     BrowserWindow.addDevToolsExtension(path.join(__dirname, 'dev', 'react-dev-tools'))
+    watchAndReload()
```
soyuka: nice, this means we can run only `npm start` in dev mode to watch?

package.json
```diff
@@ -5,27 +5,51 @@
...
-    "test:unit": "NODE_ENV=test babel-tape-runner -r ./unit-tests/_helpers/*.js unit-tests/*.js",
+    "test:unit": "cross-env NODE_ENV=test babel-tape-runner -r ./unit-tests/_helpers/*.js unit-tests/*.js",
     "lint": "prettier-standard 'app/**/*.js' 'tests/**/*.js' 'lib/**/*.js' 'unit-tests/*.js' && standard",
-    "build": "webpack",
-    "watch": "webpack --watch"
+    "deps": "cross-env DEBUG=* dependency-check . --detective precinct --entry app/index.js",
+    "clean": "git clean -dfX",
+    "burnthemall": "rm -rf ~/.electron && npm run clean && rm -f package-lock.json && npm install",
+    "update-rdt": "rm -rf dev/react-dev-tools && ced fmkadmapgofadopljbjfkapdkoienihi dev/react-dev-tools",
+    "dev": "cross-env NODE_ENV=development sh -c \"(npm run start &); npm run watch\"",
```
soyuka: shouldn't it be: `npm-run-parallel start watch`?

martinheidegger: It should, I just pushed the fixes (thanks for checking, seems like i forgot to save my edit)

PR #537, opened 2018-06-12
Graceful dat.json title update

soyuka:
https://github.com/dat-land/dat-desktop/pull/458#discussion_r194382667

Not sure if we should ignore completely the title update if the `dat.json` file doesn't exist.

### DIFF
--- a/app/actions/index.js
+++ b/app/actions/index.js
@@ -258,11 +258,15 @@ export const updateTemporaryTitleValue = title => ({
   title
 })
 
-export const updateTitle = (key, path, editValue) => async dispatch => {
+export const updateTitle = (key, path, editValue) => async (dispatch, getState) => {
   const filePath = `${path}/dat.json`
-  const blob = await readFile(filePath)
-  const metadata = { ...JSON.parse(blob), title: editValue }
-  await writeFile(filePath, JSON.stringify(metadata))
+  const metadata = { ...getState().dats[key].metadata, title: editValue }
+
+  try {
+    await writeFile(filePath, JSON.stringify(metadata))
+  } catch (error) {
+    return dispatch({ type: 'WRITE_METADATA_ERROR', key, error })
+  }
 
   dispatch({
     type: 'UPDATE_TITLE',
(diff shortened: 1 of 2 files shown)


app/reducers/index.js
```diff
@@ -1,7 +1,7 @@
 'use strict'
 
 const defaultState = {
-  dats: [],
```
martinheidegger: I don't think this can simply be changed into an object. `dats` is used as an array to show the list.

app/actions/index.js
```diff
@@ -258,11 +258,24 @@ export const updateTemporaryTitleValue = title => ({
...
-  const blob = await readFile(filePath)
-  const metadata = { ...JSON.parse(blob), title: editValue }
-  await writeFile(filePath, JSON.stringify(metadata))
+  let metadata
+
+  try {
+    metadata = await readFile(filePath)
+    metadata = JSON.parse(blob)
+  } catch (_) {
+    metadata = getState().dats[key].metadata
```
martinheidegger: I think its ambiguous to use either the file or the metadata, shouldn't this be just the metadata?

martinheidegger: Also: Theoretically the dat doesn't need to exist. Would be good to have an error here if it doesnt

app/reducers/index.js
soyuka: it's used as an object in the reducer (indexed by key if I'm not mistaken). This looked like a typo to me but I'll investigate further!

app/actions/index.js
soyuka: mhh, I though as well at first but this action is only used when changing a dat's title in the view so the assumption seemed alright!

martinheidegger:
`standard: Use JavaScript Standard Style (https://standardjs.com)
  /Users/travis/build/dat-land/dat-desktop/app/actions/index.js:270:27: 'blob' is not defined.`

soyuka:
ping @martinheidegger I've updated the PR

martinheidegger:
Hello @soyuka - Sorry for the delay, was stuck in other things.

PR #538, opened 2018-06-13
Small dev adjustments.

martinheidegger:
This PR does a few adjustments to the dev process.

- It replaces `npm-run-parallel` with `npm-run-all`. Because `npm-run-parallel` has bugs. − ceb2fe8
- It uses `optimization.nodeEnv` instead of the DefinePlugin to reduce the code. − ca58bb1


webpack.config.js
```diff
@@ -33,7 +33,7 @@ module.exports = (_, argv) => ({
   },
   plugins: [
     new DefinePlugin({
-      'process.env.NODE_ENV': argv.mode
+      'process.env.NODE_ENV': `"${argv.mode}"`
```
soyuka: why the change?

martinheidegger:
@dkastl @soyuka I spent this afternoon trying to figure out why the node-env isn't replaced. In the end it turned out that that zombie-process was rewriting the bundle. I still changed some logic here cause it seems to be more effective.

martinheidegger:
@soyuka Thanks for the approvement.

PR #551, opened 2018-10-01
Fix/clean npm scripts

dkastl:
I found the npm scripts a bit difficult to understand, not only from the point of view of the names, but also how they were organized. 

I tried to use a scope to group them together, where the "short" command is usually the one, that contains them all.

I haven't done anything with `pack` and `dist`. I didn't figure out yet, why there are 6 script entries necessary for something, that probably will be run with CI tools. I also can't think of a case, where all operating systems are built with a single run. I would suggest to reduce those to just `pack` and `dist` for the current OS.


app/components/table-row.js
```diff
@@ -142,13 +142,17 @@ const Row = ({
   const { writable, metadata, key } = dat
   const { title } = metadata
   const placeholderTitle = `#${key}`
-  return (<Tr onClick={() => inspectDat(dat.key)}>
```
martinheidegger: How does that change come by? Why is it part of this PR?

package.json
```diff
@@ -7,19 +7,19 @@
   "description": "Dat Desktop App",
   "author": "Dat Team <community@datproject.org>",
   "scripts": {
-    "start": "cross-env NODE_V=\"$(node -v)\" NODE_ENV=development electron .",
-    "pretest": "npm run lint",
-    "test": "npm run test:unit",
+    "start": "cross-env NODE_ENV=development npm-run-all -p watch start:dev",
+    "start:dev": "cross-env NODE_V=\"$(node -v)\" NODE_ENV=development electron .",
```
martinheidegger: Maybe name "start: electron"?

package.json
```diff
@@ -7,19 +7,19 @@
...
+    "watch": "webpack --watch --mode=development",
```
martinheidegger: Maybe "start:watch"

package.json
```diff
@@ -53,8 +53,8 @@
     "prettier-standard": "^8.0.1",
     "standard": "^11.0.1",
     "tape": "^4.9.0",
-    "webpack": "^4.8.3",
-    "webpack-cli": "^2.1.3",
+    "webpack": "^4.20.2",
```
martinheidegger: Why is it part of this PR?

package.json
dkastl: Yes, I had the same idea first. Will change.

package.json
dkastl: I thought changing this will make some people angry ;-)
But I agree with `start:watch`.

app/components/table-row.js
dkastl: The linter does that.

package.json
dkastl: Commands calling webpack will not run anymore with the previous version because of this issue: https://github.com/webpack/webpack/issues/8082

dkastl:
FYI, `npm install` is not run anymore automatically after `npm run clean`. I think it makes the documentation more consistent to just always have to run `npm install`.

dkastl:
@martinheidegger `npm-run-all -p start:watch start:electron` does run both processes in parallel, but if you start from scratch, it seems that ic may happen, that you first had to wait for `start:watch` to finish, because otherwise webpack would not have finished. The second time it all works.

package.json
dkastl: For the sake of https://hacktoberfest.digitalocean.com/ I should split this into more pull requests ;-)

martinheidegger:
Oh right! Yes, that is an issue. We would need a `start:build` script to be executed before `npm-run-all`

dkastl:
> Oh right! Yes, that is an issue. We would need a `start:build` script to be executed before `npm-run-all`

Can we just use `build:dev" script, because this already exists and isn't really used. `build:prod` is used for packaging and distributions. Or better rename it to `start:build`?

martinheidegger:
Right `build:dev` is fine.

martinheidegger:
Can you rebase again (with the removed package-lock.json)?

dkastl:
@martinheidegger would you mind to do this, because I don't have this cool Git software, that you have ;-)

I'm pretty much done here, so if you want to make this pull request prettier in terms of commit messages, rebasing, squashing and all that stuff, that I needed to consult a manual, please go ahead!

PR #560, opened 2018-10-16
Update/electron

AtuyL:
This PR updated `electron` from version ^2.0.1 to ^3.0.4, including `electron-builder`.

Here are the releason notes for the current electron release.
https://electronjs.org/blog/electron-3-0
https://github.com/electron/releases


dkastl:
Sorry, guys. It's the attempt to simply include `- windows` to the build matrix, that makes Travis fail now. Unfortunately it didn't work that easy to build on Windows, so feel free to ignore the failed check or just comment out https://github.com/dat-land/dat-desktop/blob/fix/react-travis/.travis.yml#L9

PR #562, opened 2018-10-17
Fix: Update inspection data only in "inspect" screen

AtuyL:
I noticed that an error with the Inspection screen.
It tried to update the inspected DAT even when navigating to other screens.

https://github.com/dat-land/dat-desktop/pull/562/files#diff-bf58091931fe43a77612c0916614753dL8

This PR now updates the dat only when the inspect screen is selected.
This PR is based on #561

### DIFF
--- a/app/consts/screen.js
+++ b/app/consts/screen.js
@@ -1,0 +1,11 @@
+export const INTRO = 'intro'
+export const DATS = 'dats'
+export const DOWNLOAD = 'download'
+export const INSPECT = 'inspect'
+
+export default {
+    INTRO,
+    DATS,
+    DOWNLOAD,
+    INSPECT
+}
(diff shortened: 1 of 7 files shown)


app/consts/screen.js
```diff
@@ -0,0 +1,11 @@
+export const INTRO = 'intro'
```
martinheidegger: Nice!

app/containers/inspect.js
```diff
@@ -1,11 +1,12 @@
 'use strict'
 
+import SCREEN from '../consts/screen'
 import Inspect from '../components/inspect'
 import { closeInspectDat } from '../actions'
 import { connect } from 'react-redux'
 
 const mapStateToProps = state => ({
-  dat: state.dats[state.inspect.key]
+  dat: state.screen === SCREEN.INSPECT ? state.dats[state.inspect.key] : null
```
martinheidegger: Nice catch!

PR #570, opened 2018-11-07
Fixed deprecated syntax in StyledComponent.

SirRujak:
The master branch was giving off a warning about the "extend" API. This replaces it with the new "styled" API.

For reference here is the original error:

```
ThemeProvider.js:28 Warning: The "extend" API will be removed in the upcoming v4.0 release. Use styled(StyledComponent) instead. You can find more information here: https://github.com/styled-components/styled-components/issues/1546
(anonymous) @ ThemeProvider.js:28
```

### DIFF
--- a/app/components/button.js
+++ b/app/components/button.js
@@ -29,7 +29,7 @@ const BaseButton = styled.button.attrs({
   }
 `
 
-const HeaderButton = BaseButton.extend`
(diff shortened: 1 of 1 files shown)


martinheidegger:
@SirRujak Thank you for the PR.

Issue #471, opened 2018-01-28
Integration tests


juliangruber:
Make sure the tests from https://github.com/datproject/dat-desktop/tree/master/tests all pass (sans dat-manager)

martinheidegger:
Landed in master with #575

PR #575, opened 2018-11-10
Integration tests

martinheidegger:
I was looking why the integration tests were missing even though #471 was merged. It seems like it was accidentally removed in 4830c335cf19fc61849e3f8e539626acc4acd20b .

When trying to enable it, I noticed that it doesn't work with Spectron because of 43acfb87d988b40aecae545f7d9e5bdc0877eeba and the separation of the browser process from the node functionality (for security reasons).

This PR fixes this by re-adding spectron to the test commands and skip the security infrastructure when running in the test context.


martinheidegger:
Seems like travis doesn't like it...?!

martinheidegger:
Thank you both for approving the PR, it still fails on Travis! Do you have an idea how to make it work?

martinheidegger:
Seems like a rerun of the travis tests did the magic ?!

PR #582, opened 2018-11-17
Performance: prevent re-rendering of icons.

AtuyL:
The DAT icons have been added through the `Sprite` component to the DOM, this way every time the App was rendered, all icons needed to be processed.

This PR adds the icons once on application start and removes them from the react rendering system.

### DIFF
--- a/app/index.js
+++ b/app/index.js
@@ -46,7 +46,11 @@ document.title = 'Dat Desktop | Welcome'
 datMiddleware
   .loadFromDisk()
   .then(function () {
+    // # addGlobalComponents
+    // Adding global components only once to the DOM.
     document.body.appendChild(datIcons())
+  })
+  .then(function () {
     render(
       <Provider store={store}>
         <App />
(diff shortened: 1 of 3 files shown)


app/index.js
```diff
@@ -45,6 +46,7 @@ document.title = 'Dat Desktop | Welcome'
 datMiddleware
   .loadFromDisk()
   .then(function () {
+    document.body.appendChild(datIcons())
```
martinheidegger: I think it might be better to have a comment here, explaining why this is outside of the react rendering.

Something like:

```javascript
// To reduce the load of the react rendering processwe simply add the icons to the dom.
```

AtuyL: @martinheidegger san
This PR have a two purpose.
1: React-Tree everything need the "CLEAN". ( ENV-Components is not required for React-Tree ! )
2: And, I think a next task is "CLEAN-ROUTING". The "Sprite" component is obstacle for that task. ( In every routing require the render a "Sprite" component. but it is waste and dirty. )

martinheidegger: How about following comment then:

```javascript
.then(function addGlobalComponents () {
  // Adding global components only once to the DOM.
  document.body.appendChild(datIcons())
})
.then(function () {
```

AtuyL: I did added.
Thank you for recommendation.

PR #585, opened 2018-11-17
Simplified Readme and Markdown syntax check

dkastl:
I cleaned up the Readme file and removed outdated (and wrong) content. 

The first paragraphs had been aligned, which requires to use HTML content in Markdown, which I would like to avoid, if possible. I prefer to use default styling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,27 +1,14 @@
-<h1 align="center">Dat Desktop</h1>
+# Dat Desktop
 
-<div align="center">
-  <strong>Peer to peer data sharing app built for humans.</strong>
-</div>
+**Peer to peer data sharing app built for humans.**
 
-<br />
-
-<div align="center">
-  <!-- Build Status -->
-  <a href="https://travis-ci.org/dat-land/dat-desktop">
-    <img src="https://img.shields.io/travis/dat-land/dat-desktop/master.svg?style=flat-square"
-      alt="Build Status" />
-  </a>
-  <!-- Standard -->
-  <a href="https://standardjs.com">
-    <img src="https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat-square"
-      alt="Standard" />
-  </a>
-</div>
+[!["Build Status"](https://img.shields.io/travis/dat-land/dat-desktop/master.svg?style=flat-square)](https://travis-ci.org/dat-land/dat-desktop) 
+[!["Standard](https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat-square)](https://standardjs.com)
 
 ![screenshot](assets/screenshot.png)
 
 ## Table of Content
+
 - [Download](#download)
 - [Commands](#commands)
 - [FAQ](#faq)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -61,10 +55,12 @@ $ npm install
 $ ELECTRON_MIRROR="https://npm.taobao.org/mirrors/electron/" npm install
 ```
 
-## License
-MIT
+## Licenses
+
+[MIT License](https://github.com/dat-land/dat-desktop/blob/master/LICENSE)
```
martinheidegger: Afaik many markdown implementations now support `[./LICENSE]` which would prevent us from having to rewrite it whenever that might change.

README.md
```diff
@@ -1,58 +1,52 @@
...
-$ npm install             # install dependencies
-$ npm start               # start the application
-$ npm run dist            # compile the app into an ASAR file to release
-$ npm start --data=<dir>  # change the path for new dat archives (default ~/Downloads/dat)
-$ npm start --db=<dir>    # change the path for users settings (default ~/.dat-desktop)
+To run _Dat Desktop_ in development mode:
+
+```sh
+npm install             # install dependencies
+npm run start           # start the application
```
martinheidegger: `npm` has the short-cut `start` → `npm start` should work.

README.md
```diff
@@ -1,58 +1,52 @@
...
 - [License](#license)
 
 ## Download
 
-To run the app locally check out our [download
-guide](https://datproject.org/install) or install directly from the command
-line using [Homebrew Cask](https://caskroom.github.io). Currently only MacOS is
-supported.
+_Dat_ needs to be installed first. To run the app locally check out our [download
+guide](https://datproject.org/install) or install directly from the command line:
```
martinheidegger: I never tried installing it through homebrew-cask. Do we have to do something to update them? Should we create issues to deploy DAT for debian/windows-store/mac-store/etc. ?

README.md
```diff
@@ -1,58 +1,52 @@
...
-    <img src="https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat-square"
-      alt="Standard" />
-  </a>
-</div>
+# Dat Desktop
+
+**Peer to peer data sharing app built for humans.**
+
+[!["Build Status"](https://img.shields.io/travis/dat-land/dat-desktop/master.svg?style=flat-square)](https://travis-ci.org/dat-land/dat-desktop) 
+[!["Standard](https://img.shields.io/badge/code%20style-standard-brightgreen.svg?style=flat-square)](https://standardjs.com)
```
martinheidegger: I am sure there were aesthetic reasons for this to be there. I am kind-of okay with this change but am not sure how design-people feel about it.

README.md
```diff
@@ -1,58 +1,52 @@
...
 - [FAQ](#faq)
 - [License](#license)
 
 ## Download
 
-To run the app locally check out our [download
-guide](https://datproject.org/install) or install directly from the command
-line using [Homebrew Cask](https://caskroom.github.io). Currently only MacOS is
-supported.
+_Dat_ needs to be installed first. To run the app locally check out our [download
```
martinheidegger: I believe the DAT homepage references https://github.com/dat-land/dat-desktop/releases when it points out how to download DAT. shouldn't we do same?

README.md
dkastl: If someone wants to change this, let them change it. In Visual Code the Markdown syntax check underlines plain HTML, and I don't like underline code ;-)

IMO this is to small thing to spend much time on.

README.md
dkastl: Any additional content is fine. I just wanted to remove, what was either outdated or didn't look necessary.

README.md
martinheidegger: Alright, let's leave it at that

PR #588, opened 2018-11-18
Removing drag-and-drop as it is incompatible with electron 3

martinheidegger:
Node 10.2 has a gyp problem which lead to `osx-mouse` not compiling properly: Effectively `electron-drag` has become not-supported. But it seems like dragging the header does work without it as well?!


AtuyL:
Can i dragging. ( Tried on Mac OS X Mojave 10.14 )

PR #596, opened 2018-11-21
Style improvements: file list

AtuyL:
This PR cleans a few style issues in the file list:

1. It puts the size at the right sight of the line.

    Before: line-right-before
    After: line-right-after

2. It fixes the `odd` line style to actually work ;). It also changes the contrast from `--color-neutral-04` to `--color-neutral-10` in order to for the lines to be better distinguishable from the background.

    | Before | Before - fixed | After |
    |---|---|---|
    | lines-before | screen shot 2018-12-03 at 11 13 48 | lines-after |

3. It removes the colon's `:` of the file titles 

    | Before | After |
    |---|---|
    | titles-before | colon-after |

### DIFF
--- a/app/components/file-list.js
+++ b/app/components/file-list.js
@@ -4,52 +4,41 @@ import React from 'react'
 import styled from 'styled-components'
 import bytes from 'prettier-bytes'
 
-const FileListContainer = styled.div`
-  :host {
-    min-height: 5rem;
-  }
-`
-
 const FileListTable = styled.table`
-  :host {
-    td {
-      padding: 0.25rem 0.5rem;
-    }
-    tr:odd td {
-      background-color: var(--color-neutral-04);
-    }
+  width: 100%;
+  border-collapse: collapse;
+  tr:odd td {
+    background-color: var(--color-neutral-04);
+  }
+  td {
+    border: 0;
+  }
+  td:last-child {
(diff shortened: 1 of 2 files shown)


app/components/file-list.js
```diff
@@ -4,52 +4,41 @@ import React from 'react'
...
-    td {
-      padding: 0.25rem 0.5rem;
-    }
-    tr:odd td {
-      background-color: var(--color-neutral-04);
-    }
+  width: 100%;
+  border-collapse: collapse;
+  tr:odd td {
+    background-color: var(--color-neutral-04);
```
martinheidegger: Shouldn't this make every other line different? In my test, each line has the same color?!

screen shot 2018-11-30 at 14 03 24

AtuyL: this styling is not working before my PR. it is other issue.

AtuyL: However, my PR was not conscious for that style.
I will work to the this issue.

AtuyL: @martinheidegger san
I FIXED. PLEASE RE-REVIEW.

AtuyL:
@martinheidegger san
> removed the colons `:`
Sorrrrrrrry !! it is another fix 😇 ( for inspect view styling )
Can i re-PR ?

PR #601, opened 2018-11-22
added: alert on add the existed dat.

AtuyL:
just title.


AtuyL:
changed: alert dialog button color.
screenshot 2018-12-05 17 40 10

PR #612, opened 2018-11-29
refactoring: moving "defaultState" moved to consts.

AtuyL:
As part of a refactoring process where I would like to clean up the react state _(it is currently polluted with problematic data)_. I extracted the `defaultState` into its own file. As part of the process I also moved it into a function to prevent changing of the data by accident.

### DIFF
--- a/app/consts/state.js
+++ b/app/consts/state.js
@@ -2,7 +2,6 @@
 
 import SCREEN from '../consts/screen'
 
-// immutable
 export default () => ({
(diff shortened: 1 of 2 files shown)


martinheidegger:
I rebased this PR and changed the naming a bit to be better understandable. The changes are available in https://github.com/dat-land/dat-desktop/tree/add/const-default-state-2 - Do you want to use those?

PR #614, opened 2018-11-30
Improve json read and write

AtuyL:
- changed to parallel writing.
   this is similar by https://github.com/dat-land/dat-desktop/commit/2a28c090269b01c77c4f19e6512083b2cc994f2c
- fix block the twice parse for json.

and minimal linting.

### DIFF
--- a/app/actions/dat-middleware.js
+++ b/app/actions/dat-middleware.js
@@ -411,13 +411,14 @@ export default class DatMiddleware {
     const datsState = Object.keys(this.dats).reduce(
       (acc, key) => ({
         ...acc,
-        [key]: JSON.stringify({
+        [key]: {
           dir: this.dats[key].path,
           opts: this.dats[key].opts
-        })
+        }
       }),
       {}
     )
+
     const pausedState = Object.keys(this.dats).reduce(
       (acc, key) => ({
         ...acc,
--- a/app/actions/dat-middleware.js
+++ b/app/actions/dat-middleware.js
@@ -393,13 +393,23 @@ export default class DatMiddleware {
     ])
 
     for (const key of Object.keys(datOpts)) {
-      const opts = datOpts[key]
-      this.internalAddDat({
-        key,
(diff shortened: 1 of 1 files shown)


app/actions/dat-middleware.js
```diff
@@ -387,19 +387,29 @@ export default class DatMiddleware {
...
-      const opts = JSON.parse(datOpts[key])
-      this.internalAddDat({
-        key,
-        path: opts.dir,
-        paused: paused[key],
-        ...opts
-      })
+      let datdata
+
+      // It is for compatibility.
```
martinheidegger:
```
// Backwards compatibility:
// In releases previous to 3.0.1 beta, the connection-data for each 
// DAT was additionally stringified. With this check we make sure
```

app/actions/dat-middleware.js
```diff
@@ -387,19 +387,29 @@ export default class DatMiddleware {
...
+      try {
+        datdata = JSON.parse(datOpts[key])
```
martinheidegger: Why did you decide to use a try-catch here? It is very slow; a `typeof` test would have been faster and it wouldn't catch other weird cases.

app/actions/dat-middleware.js
```diff
@@ -387,19 +387,29 @@ export default class DatMiddleware {
...
 
     for (const key of Object.keys(datOpts)) {
-      const opts = JSON.parse(datOpts[key])
-      this.internalAddDat({
-        key,
-        path: opts.dir,
-        paused: paused[key],
-        ...opts
-      })
+      let datdata
```
martinheidegger: Is there a reason why you renamed the variable?

AtuyL: that variable is "not opts". The `opts` have inside of.
below code from `storeOnDisk` method.

```javascript
[key]: {
    dir: this.dats[key].path,
    opts: this.dats[key].opts
}
```

PR #620, opened 2018-12-07
fix: Stopped pass the "click event" to the parentNode from DatListRow-Icons.

AtuyL:
just title 👍


okdistribute:
This was fixed in an alternative way, thanks for contributing though!

PR #632, opened 2020-02-15
fix: update Electron, comment deprecated makeSingleInstance call

aeriform-io:
There may still be a bug I am not able to identify. Clicking "Get Started" has no effect.


okdistribute:
nice thanks for this!

PR #652, opened 2020-04-08
Revert to absolute url for releases

soyuka:
/releases leads to https://github.com/datproject/dat-desktop/blob/master/releases which is a 404 I suggest reverting to the absolute url. 
Thanks for what you people do, long life to the dat protocol :D.


okdistribute:
thanks @soyuka !

